PR #274, opened 2024-10-07
Update tutorial part 1

Rich-Harris:
WIP. Effects is a TODO for now — I think we need it to be in this position, but I'm not sure what a good exercise would be (we should get rid of anything `console` based since we want to steer people towards `$inspect`). Struggling to think of examples that shouldn't just be event handlers


erickpaquin:
Hi guys. I'm not sure if I should be writing this here or not, but just wanted to let you know that in the "reactivity / state" part of the tutorial, this part doesn't look correct to me : 

{count === 1 ? 'time' : 'times'}

it should say this instead : 

{count <= 1 ? 'time' : 'times'}

Otherwise the button label prints "times" plural when the button initializes with 0.

Thanks.

erickpaquin:
My bad. I didn't know that zero was considered plural in the English language.

PR #323, opened 2024-10-10
get rid of unnecessary .json endpoints

Rich-Harris:
WIP — there's a few places where we're getting data in `load` functions via a server route. This is a hangover from the days before server `load` functions, and is entirely unnecessary


Rich-Harris:
closing as stale, we can revisit when we eventually implement PPR etc

PR #507, opened 2024-10-22
fix: also update hash after inactivity

### LINKED ISSUE #479: Playground hash should be updated before reload
At the moment if you make changes and reload the page without defocusing the editor, you'll lose work

Another option would be to just always update the hash, though IIRC there was some history-related reason we were avoiding that

dummdidumm:
closes #479 - beforeUnload was too late in my testing, and the use case is a real edge case anyway

### DIFF
--- a/apps/svelte.dev/src/routes/(authed)/playground/[id]/+page.svelte
+++ b/apps/svelte.dev/src/routes/(authed)/playground/[id]/+page.svelte
@@ -100,7 +100,9 @@
 	async function change_hash(hash?: string) {
 		let url = `${location.pathname}${location.search}`;
 		if (hash) {
-			url += `#${await compress_and_encode_text(hash)}`;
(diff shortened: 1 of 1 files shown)


Rich-Harris:
What if we updated it a) on body `mouseleave` and b) on `Cmd-R` or `Ctrl-R` `keydown` events? 5 seconds is a long time, and if it sometimes saves your work and sometimes doesn't it'll feel non-deterministic and unreliable

Conduitry:
Is `beforeUnload` too late because we need to do async stuff during it to retrieve and encode all the REPL tabs? Or is it just too late because we can't change the URL hash anymore at that point? If it's only the latter, what about a very short-lived `sessionStorage` value that we read after the reload and perform the `replaceState` then?

Rich-Harris:
merged #631 so will close this

PR #512, opened 2024-10-22
fix more safari bugs

Rich-Harris:


Rich-Harris:
confirmed this fixes some glitches

PR #519, opened 2024-10-22
chore: redirect omnisite to svelte.dev

dummdidumm:
let's see if this works correctly


Rich-Harris:
can't we just do it in a `vercel.json`? do we need any code at all?

dummdidumm:
If we don't care about the few hash-based REPLs built already then yeah that could work.

Rich-Harris:
yeah people can always copy and paste the hash themselves. closing in favour of #535

PR #531, opened 2024-10-22
Add kit.svelte.dev redirect app

benmccann:


Rich-Harris:
Rather than adding a whole SvelteKit app, what if we just had a `vercel.json` with some redirect rules? The redirects will be much quicker if the Vercel proxy can take care of them without needing to run our code

benmccann:
makes sense. according to the docs, this should work:
```
export default {
	async redirects() {
		return [
			{
				source: '/docs/:slug',
				destination: 'https://svelte.dev/docs/kit/:slug',
				permanent: true,
			},
			{
				source: '/:slug',
				destination: 'https://svelte.dev',
				permanent: true,
			}
		];
	}
};
```

though I'm not sure where to put such a thing. Do I switch to `adapter-vercel` and check it in here or do we just paste it into the Vercel web UI somewhere?

dummdidumm:
you put it in the root of the redirect app (honestly unsure if you need anything at all besides that)

Rich-Harris:
lemme try out a test project real quick

Rich-Harris:
Yeah, I just tried it out and it works, will open a PR

benmccann:
awesome. thanks!

PR #535, opened 2024-10-22
[DO NOT MERGE] chore: redirect to svelte.dev from omnisite

dummdidumm:


dummdidumm:
works flawlessly

Rich-Harris (approved):

dummdidumm:
We can't merge this one, since it would mess with the current svelte.dev deployment, so I'm in favor of #538 instead which we can also actually merge

Rich-Harris:
ah ok — will close in that case

PR #538, opened 2024-10-22
chore: setup redirects from obsolete sites

dummdidumm:
create a vercel.json for each of them and redirect to the corresponding pages on svelte.dev


apps/learn.svelte.dev/vercel.json
```diff
@@ -0,0 +1,10 @@
+{
+	"$schema": "https://openapi.vercel.sh/vercel.json",
+	"redirects": [
+		{
+			"source": "/(.*)",
```
dummdidumm, line 5: this will be cumbersome; we basically need to go through all the links and find out the right equivalent on the new one

dummdidumm:
ending my workday now, feel free to take over

apps/learn.svelte.dev/vercel.json
```diff
@@ -0,0 +1,10 @@
...
+			"destination": "https://svelte.dev/$1",
+			"permanent": true
```
benmccann, line 7: might want to skip `permanent` on whatever fallback we have for this one so that if there's some URL we missed we have an opportunity to fix it

benmccann:
should we also add a `vercel.json` for https://svelte-5-preview.vercel.app/?

dummdidumm:
Didn't get around to adding this here, but I think for that one we should have an app, so that we can redirect the hash links - there's too many of them by now to not migrate them

Rich-Harris:
i think it'll break all the hash links — might need to redirect from the browser, not sure

working on the tutorial slugs

Rich-Harris:
alright, i feel good about this — will merge and start updating the site configs

apps/kit.svelte.dev/vercel.json
```diff
@@ -0,0 +1,29 @@
+{
+	"$schema": "https://openapi.vercel.sh/vercel.json",
+	"redirects": [
+		{
+			"source": "/",
+			"destination": "https://svelte.dev/",
+			"permanent": true
+		},
+		{
+			"source": "/docs/modules",
```
benmccann, line 10: we need a URL for each individual type here

benmccann: I think we're going to have to build an app which checks what the hash is and redirects it appropriately to the correct page

PR #543, opened 2024-10-22
clarify `svelte:component` migration, avoids common gotcha

rgon:
Clarify docs to specify that, while `svelte:component` is no longer needed and deprecated, capitalizing the component variable name  is required to avoid mismatching types. 

Avoids users doing the same gotcha that I stumbled upon, it will be a common oversight when refactoring existing codebases: https://github.com/sveltejs/svelte/issues/13795

As evidenced in these docs themselves (corrected):
```svelte
...
	let component: Component<{ foo: string }> = $state(
		Math.random() ? ComponentA : ComponentB
	);
</script>

<svelte:component this={component} foo="bar" />

```

### DIFF
--- a/apps/svelte.dev/content/docs/svelte/07-misc/07-v5-migration-guide.md
+++ b/apps/svelte.dev/content/docs/svelte/07-misc/07-v5-migration-guide.md
@@ -723,6 +723,8 @@ This is no longer true in Svelte 5:
 <svelte:component this={Thing} />
 ```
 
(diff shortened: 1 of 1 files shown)


benmccann:
PRs should go to https://github.com/sveltejs/svelte/tree/main/documentation

dummdidumm:
Thank you - as Ben points out, the change should happen in the svelte repo, which will eventually be synced in here

rgon:
Thanks, moved to https://github.com/sveltejs/svelte/pull/13835

PR #548, opened 2024-10-22
fix:playground mobile overflow

TheCubiq:
Fixed an issue, that caused hiding the input/output switch on mobiles:

image

the fix: 
```css
.container {
  min-height: 0;
}
```

> By default, the container cannot be smaller than the size of its content.
> Its initial size is set to min-height: auto.

https://github.com/user-attachments/assets/f423a149-5a68-4731-9c97-f36400803053

Simple fix but hope it helps,
ly Rich <33


dummdidumm (approved):
thank you!

PR #550, opened 2024-10-22
sync

Rich-Harris:


benmccann:
This PR is going to break a whole lot of stuff if https://github.com/sveltejs/svelte.dev/issues/574 really is an issue

Rich-Harris:
I think we're fine as long as the links are to `/tutorial` and not `/tutorial/`. Just need to fix the remaining broken links though 😭

benmccann:
Now it's failing because of https://github.com/sveltejs/svelte.dev/issues/556. There are going to be easily hundreds of broken links if we don't fix that one

benmccann (approved):

PR #560, opened 2024-10-23
Fix: long <code> elements overflow mobile Safari, shows horizontal scroll bar

SkydiverTyler:
## Issue

Inside the "Svelte 5 Migration Guide" page, under the "Modern Browser Required" section, there is a list of inline `<code>` elements (starting with `clientWidth`...) that horizontally overflow on Mobile Safari (iOS 18.0.1), causing a horizontal scrollbar to appear.

## Change summary

Inside the `<Text>` component, adding the `line-break: normal;` property to the relevant CSS selector allows the inline `<code>` elements to wrap with the text as expected, eliminating the horizontal scroll bar on the page mentioned above.

https://developer.mozilla.org/en-US/docs/Web/CSS/line-break

## Screenshots

BEFORE:

IMG_3774

AFTER:

IMG_3779

### DIFF
--- a/packages/site-kit/src/lib/components/Text.svelte
+++ b/packages/site-kit/src/lib/components/Text.svelte
@@ -101,6 +101,7 @@
 
 		code:not(pre *),
 		kbd {
+			line-break: normal;
 			white-space: pre-wrap;
 			padding: 0.2rem 0.4rem;
 			margin: 0 0.2rem;


Rich-Harris:
Thank you! Unfortunately this seems to have no effect on Chrome. I've opened #577 as a potential alternative to a CSS fix, though it might be too much of an aesthetic downgrade

dummdidumm:
Closing in favor of #577

PR #562, opened 2024-10-23
docs: remove v5 warning

braebo:


dummdidumm:
Thanks! PRs for the Svelte(Kit) docs should go into those repositories, and we sync them there.

PR #563, opened 2024-10-23
fix: layout shift behind command+k search

kaevupoiss:
This fixes the layout shift behind cmd+k SearchBox when it is toggled on-off. Words in paragraphs jump and images scale.

- Added `scrollbar-gutter` to the html when `overflow` is `hidden` by the `overlay_open` store. This allows body size to remain stable. 
- Changed body positioning from `fixed` to `absolute` to account for the `scrollbar-gutter`.
- Removed `top` from frozen body to account for the new positioning.
- Added local state to the SearchBox because the last scroll position is not stored in the `top` style declaration of the body anymore.

---

`scrollbar-gutter` is not yet supported in Safari, this means the experience is unchanged for Safari users.

### DIFF
--- a/packages/site-kit/src/lib/search/SearchBox.svelte
+++ b/packages/site-kit/src/lib/search/SearchBox.svelte
@@ -26,6 +26,8 @@ It appears when the user clicks on the `Search` component or presses the corresp
 	let search: any = $state(null);
 	let recent_searches: any[] = $state([]);
 
+	let last_scroll_position: number | null = null;
+
 	let worker: Worker;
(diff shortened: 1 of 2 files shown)


Rich-Harris:
Thanks! Code looks good though I'll let someone else (@dummdidumm?) approve the PR since as a Mac user this issue doesn't affect me in the first place. (By the same token, I think it's fine that Safari doesn't support this property, because all desktop Safari users are Mac users)

dummdidumm (approved):
thanks!

Issue #307, opened 2024-10-09
Speed up build


Rich-Harris:
It takes forever to build the site, and I suspect there are some easy things we could do to speed it up but I don't know what they are (caching the twoslash transformer? caching the `Marked` instance?)

Rich-Harris:
after #713 building takes 30 seconds locally, 1m15 on Vercel. that'll do for now

PR #569, opened 2024-10-23
fix: speed up build by ensuring stable cache

dummdidumm:
Our builds got slower because the cache hash was always different. This happened because the dependencies were calculated based on the generated code after Vite compiled it, i.e. it would check stuff in `.svelte-kit/output/server/chunks/...js`. This seems to have worked fine for a while, but now Vite mashes things up such that code from `svelte.dev/src/lib/server/content.ts` is mixed in, which has loads of Vite hashes in it, and so the file content is always different. That in turn means our hash is always different, and so the snippet cache never hits.

The pragmatic solution is to hard-code the path to the original file and add a sanity check to ensure it's still accurate.

### DIFF
--- a/apps/svelte.dev/content/tutorial/03-sveltekit/02-routing/01-pages/index.md
+++ b/apps/svelte.dev/content/tutorial/03-sveltekit/02-routing/01-pages/index.md
@@ -2,7 +2,7 @@
 title: Pages
 ---
 
-SvelteKit uses x filesystem-based routing, which means that the _routes_ of your app — in other words, what the app should do when a user navigates to a particular URL — are defined by the directories in your codebase.
+SvelteKit uses filesystem-based routing, which means that the _routes_ of your app — in other words, what the app should do when a user navigates to a particular URL — are defined by the directories in your codebase.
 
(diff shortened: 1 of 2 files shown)


apps/svelte.dev/content/tutorial/03-sveltekit/02-routing/01-pages/index.md
```diff
@@ -2,7 +2,7 @@
 title: Pages
 ---
 
-SvelteKit uses filesystem-based routing, which means that the _routes_ of your app — in other words, what the app should do when a user navigates to a particular URL — are defined by the directories in your codebase.
+SvelteKit uses x filesystem-based routing, which means that the _routes_ of your app — in other words, what the app should do when a user navigates to a particular URL — are defined by the directories in your codebase.
```
Rich-Harris, line 5:
```suggestion
SvelteKit uses filesystem-based routing, which means that the _routes_ of your app — in other words, what the app should do when a user navigates to a particular URL — are defined by the directories in your codebase.
```

Rich-Harris:
edited the comment to remove 'closes' — would like to keep #307 open as even with the cache it feels like things could be a hell of a lot faster. It definitely shouldn't take over a minute to render a few pages of markdown

dummdidumm:
In my local testing it barely took 50 seconds once the cache was properly in place

PR #577, opened 2024-10-23
separate code spans

Rich-Harris:
alternative to #560. This is just a test, we would need to make the same change in the `svelte` and `kit` repos


apps/svelte.dev/content/docs/kit/20-core-concepts/30-form-actions.md
```diff
@@ -68,7 +68,7 @@ export const actions = {
 };
 ```
 
-To invoke a named action, add a query parameter with the name prefixed by a `/` character:
+To invoke a named action, add a query parameter with the name prefixed by a ` / ` character:
```
Conduitry, line 71:
```suggestion
To invoke a named action, add a query parameter with the name prefixed by a `/` character:
```

Overzealous find-replace.

apps/svelte.dev/content/docs/kit/20-core-concepts/40-page-options.md
```diff
@@ -94,7 +94,7 @@ SvelteKit will discover pages to prerender automatically, by starting at _entry
 /blog/[slug]  # dynamic, because of `[slug]`
 ```
 
-...SvelteKit will prerender `/` and `/blog`, and in the process discover links like `<a href="/blog/hello-world">` which give it new pages to prerender.
+...SvelteKit will prerender ` / ` and `/blog`, and in the process discover links like `<a href="/blog/hello-world">` which give it new pages to prerender.
```
Conduitry, line 97:
```suggestion
...SvelteKit will prerender `/` and `/blog`, and in the process discover links like `<a href="/blog/hello-world">` which give it new pages to prerender.
```

Overzealous find-replace.

apps/svelte.dev/content/docs/kit/25-build-and-deploy/55-single-page-apps.md
```diff
@@ -35,7 +35,7 @@ The `fallback` page is an HTML page created by SvelteKit from your page template
 
 On some hosts it may be `index.html` or something else entirely — consult your platform's documentation.
 
-> [!NOTE] Note that the fallback page will always contain absolute asset paths (i.e. beginning with `/` rather than `.`) regardless of the value of [`paths.relative`](configuration#paths), since it is used to respond to requests for arbitrary paths.
+> [!NOTE] Note that the fallback page will always contain absolute asset paths (i.e. beginning with ` / ` rather than `.`) regardless of the value of [`paths.relative`](configuration#paths), since it is used to respond to requests for arbitrary paths.
```
Conduitry, line 38:
```suggestion
> [!NOTE] Note that the fallback page will always contain absolute asset paths (i.e. beginning with `/` rather than `.`) regardless of the value of [`paths.relative`](configuration#paths), since it is used to respond to requests for arbitrary paths.
```

Overzealous find-replace.

apps/svelte.dev/content/docs/kit/30-advanced/10-advanced-routing.md
```diff
@@ -128,7 +128,7 @@ src/routes/[...catchall]/+page.svelte
 
 ## Encoding
 
-Some characters can't be used on the filesystem — `/` on Linux and Mac, `\ / : * ? " < > |` on Windows. The `#` and `%` characters have special meaning in URLs, and the `[ ] ( )` characters have special meaning to SvelteKit, so these also can't be used directly as part of your route.
+Some characters can't be used on the filesystem — ` / ` on Linux and Mac, `\ / : * ? " < > |` on Windows. The `#` and `%` characters have special meaning in URLs, and the `[ ] ( )` characters have special meaning to SvelteKit, so these also can't be used directly as part of your route.
```
Conduitry, line 131:
```suggestion
Some characters can't be used on the filesystem — `/` on Linux and Mac, `\ / : * ? " < > |` on Windows. The `#` and `%` characters have special meaning in URLs, and the `[ ] ( )` characters have special meaning to SvelteKit, so these also can't be used directly as part of your route.
```

Overzealous find-replace.

apps/svelte.dev/content/docs/kit/30-advanced/10-advanced-routing.md
```diff
@@ -189,7 +189,7 @@ src/routes/
 └ +layout.svelte
 ```
 
-You can also put a `+page` directly inside a `(group)`, for example if `/` should be an `(app)` or a `(marketing)` page.
+You can also put a `+page` directly inside a `(group)`, for example if ` / ` should be an `(app)` or a `(marketing)` page.
```
Conduitry, line 192:
```suggestion
You can also put a `+page` directly inside a `(group)`, for example if `/` should be an `(app)` or a `(marketing)` page.
```

Overzealous find-replace.

apps/svelte.dev/content/docs/kit/60-appendix/40-migrating.md
```diff
@@ -127,9 +127,9 @@ Previously, layout components received a `segment` prop indicating the child seg
 
 ### URLs
 
-In Sapper, all relative URLs were resolved against the base URL — usually `/`, unless the `basepath` option was used — rather than against the current page.
+In Sapper, all relative URLs were resolved against the base URL — usually ` / `, unless the `basepath` option was used — rather than against the current page.
```
Conduitry, line 130:
```suggestion
In Sapper, all relative URLs were resolved against the base URL — usually `/`, unless the `basepath` option was used — rather than against the current page.
```

Overzealous find-replace.

apps/svelte.dev/content/docs/kit/60-appendix/40-migrating.md
```diff
@@ -127,9 +127,9 @@ Previously, layout components received a `segment` prop indicating the child seg
...
 
-This caused problems and is no longer the case in SvelteKit. Instead, relative URLs are resolved against the current page (or the destination page, for `fetch` URLs in `load` functions) instead. In most cases, it's easier to use root-relative (i.e. starts with `/`) URLs, since their meaning is not context-dependent.
+This caused problems and is no longer the case in SvelteKit. Instead, relative URLs are resolved against the current page (or the destination page, for `fetch` URLs in `load` functions) instead. In most cases, it's easier to use root-relative (i.e. starts with ` / `) URLs, since their meaning is not context-dependent.
```
Conduitry, line 132:
```suggestion
This caused problems and is no longer the case in SvelteKit. Instead, relative URLs are resolved against the current page (or the destination page, for `fetch` URLs in `load` functions) instead. In most cases, it's easier to use root-relative (i.e. starts with `/`) URLs, since their meaning is not context-dependent.
```

Overzealous find-replace.

apps/svelte.dev/content/docs/kit/98-reference/50-configuration.md
```diff
@@ -773,7 +773,7 @@ base?: '' | `/${string}`;
...
+A root-relative path that must start, but not end with ` / ` (e.g. `/base-path`), unless it is the empty string. This specifies where your app is served from and allows the app to live on a non-root path. Note that you need to prepend all your root-relative links with the base value or they will point to the root of your domain, not your `base` (this is how the browser works). You can use [`base` from `$app/paths`](https://kit.svelte.dev/docs/modules#$app-paths-base) for that: `<a href="{base}/your-page">Link</a>`. If you find yourself writing this often, it may make sense to extract this into a reusable component.
```
Conduitry, line 776:
```suggestion
A root-relative path that must start, but not end with `/` (e.g. `/base-path`), unless it is the empty string. This specifies where your app is served from and allows the app to live on a non-root path. Note that you need to prepend all your root-relative links with the base value or they will point to the root of your domain, not your `base` (this is how the browser works). You can use [`base` from `$app/paths`](https://kit.svelte.dev/docs/modules#$app-paths-base) for that: `<a href="{base}/your-page">Link</a>`. If you find yourself writing this often, it may make sense to extract this into a reusable component.
```

Overzealous find-replace.

apps/svelte.dev/content/docs/kit/30-advanced/40-service-workers.md
```diff
@@ -64,7 +64,7 @@ self.addEventListener('fetch', (event) => {
 		const url = new URL(event.request.url);
 		const cache = await caches.open(CACHE);
 
-		// `build`/`files` can always be served from the cache
+		// `build` / `files` can always be served from the cache
```
Conduitry, line 67: Presumably not strictly necessary here. I don't know whether this still makes sense to do anyway.

Rich-Harris: I think we may as well be consistent

dummdidumm:
can we automate this somehow? I feel like we're going to make this mistake over and over again if we're not super careful

Conduitry:
Automating this via the renderer sounds good, if we can make it work properly. As Rich's initial mass-find-replace indicated, there are a lot of false positives if you're just looking for those three specific characters in a row. But maybe it might more doable in a renderer if we have access to more AST-y stuff.

Rich-Harris:
The first question to answer is whether we like this change:

image

image


Personally I think it's actually an improvement, but want to be sure others agree. 

As for automating: are we imagining a warning or actually transforming it? Not sure off the top of my head if there's a single place to put that logic

dummdidumm:
Actually transform it.
Agree it looks better

apps/svelte.dev/content/docs/kit/20-core-concepts/10-routing.md
```diff
@@ -170,7 +170,7 @@ To create a layout that applies to every page, make a file called `src/routes/+l
 <slot></slot>
 ```
 
-If we create pages for `/`, `/about` and `/settings`...
+If we create pages for ` / `, `/about` and `/settings`...
```
Conduitry, line 173:
```suggestion
If we create pages for `/`, `/about` and `/settings`...
```

packages/site-kit/src/lib/markdown/renderer.ts
```diff
@@ -283,6 +283,24 @@ export async function render_content_markdown(
...
+				// @ts-expect-error
+				const before = token.tokens[slash_index - 1];
+				// @ts-expect-error
+				const after = token.tokens[slash_index + 1];
+
+				if (before?.type === 'codespan' && after?.type === 'codespan') {
+					// @ts-expect-error
+					token.tokens[slash_index].raw = token.tokens[slash_index].text = ' / ';
+				}
+			}
```
Conduitry, line 303: Do we actually want to check in all these changes if we're doing this transformation?

Rich-Harris: good point

.prettierignore
```diff
@@ -5,4 +5,5 @@ dist
 pnpm-lock.yaml
 .vercel
 _generated.md
-land-110m.json
\ No newline at end of file
+land-110m.json
+/apps/svelte.dev/content/docs
```
Conduitry, line 9: We should be able to revert this too. I'm assuming this was done because Prettier had other opinions about what you were trying to check in?

Rich-Harris: it did. i think it'll save us bother in future, so i figured we may as well keep it

dummdidumm (approved):

PR #580, opened 2024-10-23
chore: Fix argument name in inline event handler

drewyoungren:
In [Basic Svelte / Events / inline handlers](https://svelte.dev/tutorial/svelte/inline-handlers):

Event handler has incorrect argument. Works as is because of (deprecated) `Window.event` (I think).


dummdidumm (approved):
thank you!

PR #584, opened 2024-10-23
docs: fix pointer move event parameter for consistency

banu-teja:
- Changed event parameter from 'e' to 'event' in onpointermove handler for consistency with instructions.

https://svelte.dev/tutorial/svelte/inline-handlers


dummdidumm (approved):
thank you!

PR #585, opened 2024-10-23
docs: remove hidden overlay on snippet controls

ivenuss:
## Problem
Snippet `controls` element without the `filename` children overlays the actual snippet code and the first line becomes hard to select and downgrades the UX.

This happens because `controls` div has full width and `z-index: 2` which makes content under it hardly selectable.

image

## Solution
Remove the 100% width and make the width auto according to the contents of its children + add the `right: 0` position to be placed in the top right corner. 

This solution works for both types:
- Without filename image
- With filename (width grows automatically) image


Rich-Harris:
thanks!

Issue #556, opened 2024-10-22
lowercase hash links?


benmccann:
https://svelte.dev/docs/kit/types#generated-types turned into https://svelte.dev/docs/kit/types#Generated-types, which I think both looks worse and is unnecessarily breaking

Though it's a bit nicer here: https://svelte.dev/docs/kit/$app-navigation#pushState. So I think I'd lowercase only for titles and not for hashes generated from reference docs if possible

What we decide here may affect https://github.com/sveltejs/kit/pull/12862

benmccann:
I think the way to do this might be to put code references in backticks when they are in the headings. We could then lowercase everything that is not in backticks. We shouldn't change the styling for backticks in headings though

PR #586, opened 2024-10-23
fix: redirect hashes with different casing

benmccann:
Closes https://github.com/sveltejs/svelte.dev/issues/556. This works pretty well.

Is it okay that I mutated the URL from the store on line 40?
And is using goto on line 49 the appropriate way to do this?


apps/svelte.dev/src/routes/docs/[...path]/+page.svelte
```diff
@@ -17,6 +20,37 @@
...
+		// e.g. https://svelte.dev/docs/kit/@sveltejs-kit#redirect vs https://svelte.dev/docs/kit/@sveltejs-kit#Redirect
+		for (const el of elements) {
+			if (el.id === hash) {
+				return;
+			}
+		}
+		for (const el of elements) {
+			if (el.id.toLocaleLowerCase() === hash.toLocaleLowerCase()) {
+				const url = $page.url;
+				url.hash = el.id;
```
Comment (dummdidumm) [inline apps/svelte.dev/src/routes/docs/[...path]/+page.svelte:42]: is it necessary to mutate it? wouldn't you be able to construct the new url and return it without touching the store?

apps/svelte.dev/src/routes/docs/[...path]/+page.svelte
```diff
@@ -17,6 +20,37 @@
...
+	// hash was lowercase in v4 docs and varying case in v5 docs
+	function get_url_to_redirect_to() {
+		const hash = $page.url.hash.replace(/^#/i, '');
+		if (!hash) return;
+
+		const elements = document.querySelectorAll('*[id]');
+		// if there's an exact match, use that. no need to redirect
+		// also handles the case where one appears twice with difference casing
+		// e.g. https://svelte.dev/docs/kit/@sveltejs-kit#redirect vs https://svelte.dev/docs/kit/@sveltejs-kit#Redirect
+		for (const el of elements) {
```
Comment (Rich-Harris) [inline apps/svelte.dev/src/routes/docs/[...path]/+page.svelte:34]: why are we iterating? just use a case-insensitive selector

```js
const heading = document.querySelector(`[id="${id}" i]`);
```

apps/svelte.dev/src/routes/docs/[...path]/+page.svelte
Comment (benmccann) [reply apps/svelte.dev/src/routes/docs/[...path]/+page.svelte:42]: yeah, mostly I wanted to know if I should even be able to. It doesn't seem like I should. I filed a feature request for that: https://github.com/sveltejs/kit/issues/12870. And have updated this to make a copy

benmccann:
> So here's the thing: this won't close https://github.com/sveltejs/svelte.dev/issues/556, unless you also add a handleMissingId function to svelte.config.js that allows broken hashes through. Which we obviously don't want.
I think we should just fix the links.

I fixed the links in our own docs already. What I'm worried about is breaking every link on the internet, which is what I'm trying to prevent here

apps/svelte.dev/src/routes/docs/[...path]/+page.svelte
Comment (benmccann) [reply apps/svelte.dev/src/routes/docs/[...path]/+page.svelte:34]: updated to use that, but we still need to iterate. e.g. we have https://svelte.dev/docs/kit/@sveltejs-kit#redirect and https://svelte.dev/docs/kit/@sveltejs-kit#Redirect

Comment (Rich-Harris) [reply apps/svelte.dev/src/routes/docs/[...path]/+page.svelte:34]: could do this!

```js
document.querySelector(`[id="${id}"]`) ?? document.querySelector(`[id="${id}" i]`);
```

Comment (benmccann) [reply apps/svelte.dev/src/routes/docs/[...path]/+page.svelte:34]: well it turns out that this issue isn't fixable. we need to change one of these ids so that they don't collide: https://github.com/sveltejs/svelte.dev/issues/590

apps/svelte.dev/src/routes/docs/[...path]/+page.svelte
```diff
@@ -17,6 +20,35 @@
...
+	// hash was lowercase in v4 docs and varying case in v5 docs
+	function get_url_to_redirect_to() {
+		const hash = $page.url.hash.replace(/^#/i, '');
+		if (!hash) return;
+
+		// if there's an exact match, use that. no need to redirect
+		// also semi-handles the case where one appears twice with difference casing
+		// e.g. https://svelte.dev/docs/kit/@sveltejs-kit#redirect vs https://svelte.dev/docs/kit/@sveltejs-kit#Redirect
+		// but browsers make it impossible to really do: https://github.com/sveltejs/svelte.dev/issues/590
+		if (document.querySelector(`[id="${hash}"]`).length) {
```
Comment (Rich-Harris) [inline apps/svelte.dev/src/routes/docs/[...path]/+page.svelte:34]:
```suggestion
		if (document.querySelector(`[id="${hash}"]`)) {
```

apps/svelte.dev/src/routes/docs/[...path]/+page.svelte
```diff
@@ -17,6 +20,35 @@
 		const link = 'docs/' + data.document.file.split('/').slice(2).join('/');
 		return `https://github.com/sveltejs/${name}/edit/main/documentation/${link}`;
 	});
+
+	// make hash case-insensitive
+	// hash was lowercase in v4 docs and varying case in v5 docs
+	function get_url_to_redirect_to() {
+		const hash = $page.url.hash.replace(/^#/i, '');
```
Comment (Rich-Harris) [inline apps/svelte.dev/src/routes/docs/[...path]/+page.svelte:27]:
```suggestion
		const hash = $page.url.hash.slice(1);
```

apps/svelte.dev/src/routes/docs/[...path]/+page.svelte
```diff
@@ -17,6 +20,35 @@
...
+		if (!hash) return;
```
Comment (Rich-Harris) [inline apps/svelte.dev/src/routes/docs/[...path]/+page.svelte:28]:
```suggestion
		if (hash === '') return;
```

Rich-Harris (approved):

PR #593, opened 2024-10-23
fix: don't update view when it was the source of the state change

### LINKED ISSUE #484: Playground undo/redo kinda broken
- Paste a block of code multiple times
- Ctrl + Z multiple times

The second undo will undo the undo, i.e. redo. From that point it's just an undo/redo loop.

Additionally, Ctrl + Y does not appear to do anything.

### LINKED ISSUE #582: Tutorial Site Jumps around when typing
### Describe the bug

As I went through the tutorial for the new Svelte 5 tutorial, there are certain lessons that when you are typing in the editor, the line you are typing on jumps out of view while you cursor still remains on the line you are typing on. My intuition is that as certain things are typed (perhaps state related code) the preview window is reactively reloading (which is the expected behavior) but I think this might be causing the editor to jump around, which makes you have to either finish typing you code blindly or you have to continuously scroll the editor back to the line you are typing on. btw...The new tutorial (content wise) is fantastic!

### Reproduction

https://svelte.dev/tutorial/svelte/actions

https://github.com/user-attachments/assets/841f1ef1-234b-43ab-a909-680749103a3d

### Logs

_No response_

### System Info

```shell
browsers Tested: 
Arc: Version 1.63.1 (54714) Chromium Engine Version 129.0.6668.101 Mac OS
Chrome: Version 129.0.6668.103 (Official Build) (arm64) : Mac Os
Safari: Version 18.0 (20619.1.26.31.6)
```

### Severity

annoyance

dummdidumm:
When the view emits and update event, the state was updated, which in turn updated the view again. That messes with things, for example the scroll positioning. This fixes that by adding a circuit breaker. Fixes #582 and fixes #484

### DIFF
--- a/packages/editor/src/lib/Workspace.svelte.ts
+++ b/packages/editor/src/lib/Workspace.svelte.ts
@@ -295,7 +295,7 @@ export class Workspace {
 			const file = files.find((file) => file.name === name) as File;
 
 			if (file) {
-				this.#update_state(file, state);
+				this.#update_state(file, state, true);
 			} else {
 				this.states.delete(name);
 			}
@@ -314,7 +314,7 @@ export class Workspace {
 		this.#reset_diagnostics();
 	}
 
-	update_file(file: File) {
+	update_file(file: File, update_view = true) {
 		if (file.name === this.#current.name) {
 			this.#current = file;
 		}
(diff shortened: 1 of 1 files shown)


Rich-Harris:
I'd rather we avoid exposing a public option like that — opened #597 as an alternative

Rich-Harris:
merged #597, closing

PR #594, opened 2024-10-23
fix: make tutorial sections more visible

dummdidumm:
the default grayed-out styling makes it super hard to quickly find what you're looking for. This gives them a black and bold styling, which better fits their use case of separating exercises into sections.


Rich-Harris:
What is this changing? It still looks like this for me:

image

Honestly, I think we might be better off using a `<details>` anyway — the `<select>` was a quick thing to implement but I find it barely usable not to mention fugly (and FWIW it's not progressively enhanced)

dummdidumm:
Sounds like another Firefox thing where this can't be styled... FFS every browser is shit in its own way.

Issue #540, opened 2024-10-22
Redirect old docs


benmccann:
~~This returns a 404: https://svelte.dev/faq~~
~~As does this: https://svelte.dev/docs/typescript#enhancing-built-in-dom-type~~

We should probably improve some of the best-effort redirects. In https://github.com/sveltejs/svelte.dev/pull/595 I commented some that are not good now. Perhaps we can use client-side hash redirects to improve things
And the types/modules pages are wrong: https://github.com/sveltejs/svelte.dev/pull/538/files#r1811558592 (see https://github.com/sveltejs/kit/pull/12862 to manually fix them in the kit docs)

https://github.com/sveltejs/svelte.dev/issues/556 made hashes case-insensitive, but there are still places where punctuation or other things changed. E.g. https://svelte.dev/docs/kit/form-actions#progressive-enhancement-use-enhance has a punctuation change

~~We also need to review all the learn.svelte.dev redirects~~ (https://github.com/sveltejs/svelte.dev/issues/619)

benmccann:
The old `types` page gets redirected to https://svelte.dev/docs/kit/@sveltejs-kit, which is largely correct. The `public-types-` and `private-types-` prefixes could be stripped and capitalization is different.

https://kit.svelte.dev/docs/types#public-types-loadevent
https://kit.svelte.dev/docs/types#private-types-cspdirectives

The one exception is that the `Config` and `KitConfig` types now live on https://svelte.dev/docs/kit/configuration. But the old docs for those were so useless that I doubt anyone linked to them and I'm not too worried about redirecting them

benmccann:
The old `modules` page is another story. It needs a placeholder redirection page. E.g. https://kit.svelte.dev/docs/modules#$app-forms-deserialize now goes to https://svelte.dev/docs/kit/$app-forms#deserialize

PR #595, opened 2024-10-23
fix: add more svelte.dev redirects

benmccann:
continuing work on https://github.com/sveltejs/svelte.dev/issues/540

It seems I duplicated some existing redirects, but I'm not sure which ones or where they live

https://svelte.dev/docs/svelte-easing already redirects, but https://svelte.dev/docs/typescript doesn't. Any idea where the first redirect lives?

### DIFF
--- a/apps/svelte.dev/src/hooks.server.js
+++ b/apps/svelte.dev/src/hooks.server.js
@@ -23,7 +23,10 @@ const mappings = new Map([
 	['/docs/svelte-components', '/docs/svelte/svelte-files'],
 	['/docs/svelte-easing', '/docs/svelte/svelte-easing'],
 	['/docs/svelte-motion', '/docs/svelte/svelte-motion'],
-	['/docs/svelte-register', 'https://github.com/sveltejs/svelte/blob/svelte-4/documentation/docs/06-legacy/01-svelte-register.md'],
+	[
+		'/docs/svelte-register',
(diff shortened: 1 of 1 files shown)


Rich-Harris:
here https://github.com/sveltejs/svelte.dev/blob/4e429de50a0364fe7aa719230a9a6bbdae46d244/apps/svelte.dev/src/routes/docs/%5B...path%5D/%2Blayout.server.ts#L11-L18

Rich-Harris:
(I don't understand why the second link isn't redirecting)

Rich-Harris:
(it does redirect in dev)

benmccann:
I don't know either, but there are others that are currently broken despite retaining the same name:
- https://svelte.dev/docs/basic-markup
- https://svelte.dev/docs/svelte/svelte-components

I was going to say maybe just the reference docs work, but v4-migration-guide seems to work. It's really strange. I think I'll just move the layout redirects since they don't seem to be working

benmccann:
Although I suppose leaving the layout redirect there doesn't hurt just in case there's something I overlooked

Rich-Harris:
I think I understand what's happening — the redirects need to be prerendered, and the `/docs/typescript` page isn't but `/docs/svelte-easing` is. The real mystery is why some redirects _are_ being prerendered

Rich-Harris (approved):

PR #596, opened 2024-10-23
fix: colorful testimonials for hoverless devices

### LINKED ISSUE #578: Make screenshots on home page colored on mobile
You can't hover over them, so it looks more like a memorial than a warming invite

dummdidumm:
fixes #578

worked when I tried it locally with `hover: hover`, so I'm assuming it works for the opposite on mobile


Rich-Harris:
for future reference, both chrome and firefox simulate this mode when you turn on mobile view in devtools. (safari probably does too but i can't make sense of the icons in their devtools so i don't know how to turn on mobile view, nor do i care)

Rich-Harris (approved):

PR #597, opened 2024-10-23
separate state updates from file updates

Rich-Harris:
alternative to #593 that avoids adding internal implementation details to the `Workspace` public API


Rich-Harris:
self-merging since the bug is quite annoying

Issue #575, opened 2024-10-23
buttons in code blocks need a background


Rich-Harris:
<img width="359" alt="image" src="https://github.com/user-attachments/assets/7aa2b371-546a-4169-842e-ab5cb5549b23">

easy fix would be to add `background: inherit` on the `button` or `.raised` selector in the global styles, though it would probably look better if the space immediately around the button also obscured the text

cowboycodr:
When experimenting with this I accidentally entered in the wrong background color. I kind of dig it. Thoughts?

![image](https://github.com/user-attachments/assets/6dd190ef-a72d-46c8-a155-b42f31fb793f)
![image](https://github.com/user-attachments/assets/27204703-3683-46f8-8a33-38b1e2cb4622)

cowboycodr:
Just made a pull request. Went with the second option, keeping it consistent.

cowboycodr:
Should be good to close with #601.

PR #601, opened 2024-10-23
fix: add background color to code block buttons

cowboycodr:

Solves #575. Simply added a background color to the code block component, and added `background-color: inherit` to the global `.raised` class.

image


Rich-Harris:
Thanks — I put the background on `.controls` instead, to prevent text peeking out behind the right edge

Rich-Harris (approved):

PR #602, opened 2024-10-23
chore: update links to learn.svelte.dev

benmccann:
while most these would be redirected, this is likely better for SEO. it also fixes a broken link, which alerted me to another set of redirects I need to double check in a follow up


Conduitry:
Hm. I have really mixed feelings about rewriting history in blog posts. I don't know what to do about this. https://svelte.dev/blog/svelte-4, for example, contrasts learn.svelte.dev with the old tutorial that's still available for certain browsers. Do we update that too now that that distinction no longer exists?

benmccann:
> Hm. I have really mixed feelings about rewriting history in blog posts. I don't know what to do about this. https://svelte.dev/blog/svelte-4, for example, contrasts learn.svelte.dev with the old tutorial that's still available for certain browsers. Do we update that too now that that distinction no longer exists?

Yeah. I think blog posts have multiple functions. I think this helps with SEO and helps with folks who stumble upon these blog posts now trying to learn how to do something. It might hurt for folks who are trying to act as historians and for some reason needed to know that the tutorial used to live on a separate domain (though I did leave some of those references alone). I'm prioritizing the needs of current users over historians here

benmccann:
I fixed the merge conflicts on this. I tried to be judicious about which links and text to update. Happy to revert a couple more if you think that's best. There are a couple in the last post that would otherwise be broken that I definitely think we should update

Rich-Harris:
Isn't this what redirects are for?

benmccann:
I expect there's a negative SEO impact to passing on page rank if you route it through a redirected subdomain.

Plus, we have a lot of broken and missing redirects right now: https://github.com/sveltejs/svelte.dev/issues/619

Rich-Harris:
I wouldn't worry too much about SEO — if you google 'svelte tutorial' then the top result is the correct one. And it's not like these specific links are broken in the first place

benmccann:
The links in apps/svelte.dev/content/blog/2023-09-20-runes.md are broken 

We don't really have anything preventing us from breaking redirects, but if we write these as paths instead of fully qualified URLs then we get the benefit of the link checking plus we've rewritten probably several hundred URLs at this point and these are like the last dozen remaining

apps/svelte.dev/content/blog/2022-12-14-announcing-sveltekit-1.0.md
```diff
@@ -88,7 +88,7 @@ Back in early 2021, when we adopted Vite, we were the first major application fr
 
 We’ve had terrific support from teams at [Vercel](https://vercel.com), [Netlify](https://netlify.app/) and [Cloudflare](https://www.cloudflare.com/) to make zero-config deployments for those platforms possible.
 
-Our friends at [StackBlitz](https://stackblitz.com/) worked feverishly to make [learn.svelte.dev](https://learn.svelte.dev), our first-of-its-kind [WebContainer](https://blog.stackblitz.com/posts/introducing-webcontainers/)-powered interactive tutorial, a reality.
+Our friends at [StackBlitz](https://stackblitz.com/) worked feverishly to make [our new tutorial](/tutorial), the first-of-its-kind [WebContainer](https://blog.stackblitz.com/posts/introducing-webcontainers/)-powered interactive tutorial, a reality.
```
Rich-Harris, line 91: let's at least revert this, because this _is_ messing with history in a way that feels wrong. like, `learn.svelte.dev` appears in talks and blog posts

```suggestion
Our friends at [StackBlitz](https://stackblitz.com/) worked feverishly to make [learn.svelte.dev](https://learn.svelte.dev), our first-of-its-kind [WebContainer](https://blog.stackblitz.com/posts/introducing-webcontainers/)-powered interactive tutorial, a reality.
```

Rich-Harris: (and if we're reverting that, we should probably revert the other mention also in this document)

apps/svelte.dev/content/blog/2022-12-14-announcing-sveltekit-1.0.md
```diff
@@ -9,7 +9,7 @@ After two years in development, [SvelteKit](https://kit.svelte.dev) has finally
 
 We’re so excited to share this release with you. It’s the culmination of thousands of hours of work, both from the Svelte core team and the wider community, and we think it’s the most enjoyable way to build production-grade websites, whether you’re a solo developer working on a small project or part of a large team.
 
-To get started, run `npm create svelte@latest`, and visit the [docs](/docs/kit) and (experimental!) [interactive tutorial](https://learn.svelte.dev).
+To get started, run `npm create svelte@latest`, and visit the [docs](/docs/kit) and (experimental!) [interactive tutorial](/tutorial).
```
benmccann, line 12:
```suggestion
To get started, run `npm create svelte@latest`, and visit the [docs](/docs/kit) and (experimental!) [interactive tutorial](https://learn.svelte.dev).
```

apps/svelte.dev/content/blog/2022-12-14-announcing-sveltekit-1.0.md
benmccann: done. I'm fine reverting all of them honestly except for the last two which would be otherwise broken

Rich-Harris (approved):

PR #604, opened 2024-10-24
font switcher

Rich-Harris:
WIP. Need a better icon, and some of the variables could probably use more tweaking


Rich-Harris:
(also system-ui probably isn't the best sans option)

packages/site-kit/src/lib/styles/tokens.css
```diff
@@ -178,6 +178,24 @@
...
+		code {
+			font-size: 0.75em;
+		}
+	}
+
+	&.font-accessible {
+		--sk-font-family-body: 'OpenDyslexic';
+		--sk-font-size-body: 1.6rem;
+
+		code {
```
dummdidumm, line 195: So that you don't need a magnifying glass for the code blocks
```suggestion
		p code {
			font-size: 0.75em;
		}
	}

	&.font-accessible {
		--sk-font-family-body: 'OpenDyslexic';
		--sk-font-size-body: 1.6rem;

		p code {
```

packages/site-kit/src/lib/components/FontDropdown.svelte
```diff
@@ -0,0 +1,76 @@
+<script lang="ts">
+	import Dropdown from './Dropdown.svelte';
+	import Icon from './Icon.svelte';
+
+	let { top }: { top: boolean } = $props();
```
dummdidumm, line 5:
```suggestion
	let { top = false }: { top?: boolean } = $props();
```

kvetoslavnovak:
Shouldn't the toggle icon be reversed? When we're on the serif font page, the switcher should display the sans-serif icon, and when we're on the sans-serif page, it should show the serif icon. But I see the same 'reverse psychology' is used for the dark/light theme button as well! :-)

Rich-Harris:
good point! #718

Issue #617, opened 2024-10-24
Add hackathon banner to the docs


bennymi:
Currently the only people aware of the hackathon are the ones that watched the Svelte summit all the way to the end, or the ones who stumbled upon the hackathon discord channel. It would be good to have it displayed on the documentation site so more people will hear about it, as well as maybe have a proper announcement for it on the discord server.

PR #618, opened 2024-10-24
banner: Svelte Hack 2024

PuruVJ:
Closes https://github.com/sveltejs/svelte.dev/issues/617


apps/svelte.dev/src/routes/banner.json/+server.js
```diff
@@ -9,29 +9,15 @@ export const GET = async () => {
...
-				start: new Date('1 Dec, 2023 00:00:00 UTC'),
-				end: new Date('24 Dec, 2023 23:59:59 UTC'),
+				id: 'sveltehack2024',
+				start: new Date('22 Oct, 2024 00:00:00 UTC'),
+				end: new Date('10 January, 2025 23:59:59 UTC'),
 				arrow: true,
 				content: {
-					lg: 'Advent of Svelte 2023 is here!',
-					sm: 'Advent of Svelte'
+					lg: 'Svelte Hack 2024 (Runes edition) is here!',
```
dummdidumm, line 17:
```suggestion
					lg: 'Celebrating the release of Svelte 5 with Svelte Hack 2024!',
```

Rich-Harris: This reads oddly. Who is celebrating? It should probably be either "We’re celebrating the release" or "Celebrate the release" if we're going with this language. It should probably be more CTA-ish anyway though — something like this?

> Cast runes, win prizes: SvelteHack 2024

Most usages on the site are 'SvelteHack' (though the `<title>` is 'Svelte Hack' and there's at least one 'Sveltehack'). We need to ensure consistency.

apps/svelte.dev/src/routes/banner.json/+server.js
```diff
@@ -9,29 +9,15 @@ export const GET = async () => {
 	return json(
 		defineBanner([
 			{
-				id: 'advent2023',
-				start: new Date('1 Dec, 2023 00:00:00 UTC'),
-				end: new Date('24 Dec, 2023 23:59:59 UTC'),
+				id: 'sveltehack2024',
+				start: new Date('22 Oct, 2024 00:00:00 UTC'),
+				end: new Date('10 January, 2025 23:59:59 UTC'),
```
Rich-Harris, line 14: do we really want it to be visible for the entire duration?

Rich-Harris:
Things I'm working on:

- the contrast between the background and foreground is way too low
- the button should look like a button (I have a local branch that adds a `.primary` variant of the `.raised` class which we can likely use here, will try and push it up soon)
- we should get rid of the `banner.json` indirection, it's very unnecessary now that everything is in the same place.

apps/svelte.dev/src/routes/banner.json/+server.js
dummdidumm: probably makes sense to put it down mid december, you wouldn't start stuff then.

once you close it won't reappear, if that's your concern of it being annoying.

Rich-Harris:
re 03be7d9f9490b37fabe0dd0115eb15b01e13db0d, for future reference we should almost _never_ set a `font-size` (or `font-weight`, etc) individually, we should always use one of the `font` tokens in `tokens.css`:

```css
font: var(--sk-font-ui-medium);
```

The same goes for colours etc, though there's still a few places where we're breaking that rule

apps/svelte.dev/src/routes/banner.json/+server.js
dummdidumm: I wanted to put "Svelte 5" in in case someone didn't realize that already. Though I guess it becomes obvious once you click the link ... yeah I think I like your suggestion.

bleucitron:
Is it possible to handle banners without `endDate` ? i'd like to add one on the french version to warn people the translation is WIP, but i need to specify and `endDate` to make it appear.

I can set some random date in the future of course, but this could be nice to have.

bleucitron:
also, do banners need to always be links ?

(i can open an issue for these 2 points if you'd like)

PuruVJ:
> Is it possible to handle banners without endDate ? i'd like to add one on the french version to warn people the translation is WIP, but i need to specify and endDate to make it appear.

You could add a date for 100 years after now 😄. It can be added to the site, but I see Rich removed the `defineBanner` function, which couldve dealt with an infinite `endDate`, so IG endDate is now explicit.

> also, do banners need to always be links ?
You could apply an action in your own repo to the Banner component. Allowing it to do more things could make the API very complicated for not much ROI

Rich-Harris (approved):

Issue #536, opened 2024-10-22
Remove old stuff from https://github.com/sveltejs/sites


benmccann:
- everything in the `packages` directory can be removed
- readme should be updated
- should `db` directory be moved here?

PR #628, opened 2024-10-24
chore: move DB stuff over from sites repo

benmccann:
ref https://github.com/sveltejs/svelte.dev/issues/536. and also mentioned by french translation folks

Am I correct in assuming this should move here and is still used by the new site?

### DIFF
--- a/db/README.md
+++ b/db/README.md
@@ -1,0 +1,8 @@
+# Database setup
+
+Login to [Supabase](https://supabase.com) and create a database. Once done, you should be on your database's dashboard. Duplicate the `.env.example` file and rename it to `.env.local`, and set these environment variables:
+
+- `SUPABASE_URL`: The config URL
+- `SUPABASE_KEY`: The public API key
+
+Then, navigate to your database's "SQL Editor", click on "New query", and paste in [setup.sql](./setup.sql). Run this SQL to seed the database and you're good to go.
--- a/db/setup.sql
+++ b/db/setup.sql
@@ -1,0 +1,183 @@
+-- drop everything
+alter table if exists public.gist drop constraint if exists gist_userid_fkey;
+alter table if exists public.session drop constraint if exists session_userid_fkey;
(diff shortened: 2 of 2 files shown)


dummdidumm:
Should we put this stuff inside a big `contributing.md` file?

Rich-Harris:
Feels like this should go in `apps/svelte.dev`, no?

Are these instructions sufficient? Presumably you also need a GitHub app set up for auth, otherwise you can't save to the db anyway

benmccann:
> Should we put this stuff inside a big contributing.md file?

Done

> Feels like this should go in apps/svelte.dev, no?

Done

> Are these instructions sufficient? Presumably you also need a GitHub app set up for auth, otherwise you can't save to the db anyway

No idea. There's at least less missing parts now than before. But I don't really have this stuff setup or know enough to document it fully

Rich-Harris:
I don't think there's much point merging a PR until we've validated that it works

benmccann:
My main concern was trying to cleanup the sites repo without losing anything

SrGeneroso:
would be possible to add the playground as an iframe from svelte.dev so the translation sites doesn't have to host their own db?
Since the playground (can i keep calling it repl?) need the github auth and a separate db I think that would fragmentate userbase. Would be nice to have all the users in the same db. Would that be very complicated?
https://sveltefr.dev/playground/hello-world?version=5.1.3

Rich-Harris:
Stuff like this is why we'd prefer to internationalise svelte.dev than have a bunch of different sites with translations

benmccann:
I deleted all files from the sites repo except for the HN example, so these files only live here. It is impossible for me to do anything else in this PR because I don't know what the rest of the setup steps are, so I can't document them. I'd be happy if someone would like to complete this PR. Otherwise, I think this is still worth merging as some help is better than none and folks aren't going to know to look in the PR queue to find the files that are necessary to setup the DB

PR #631, opened 2024-10-25
update hash on mouseleave and before reload

Rich-Harris:
This updates the playground URL hash on `mouseleave`, which means that if you move the mouse out of the editor to select the URL bar, the hash is updated before you select it which feels a bit nicer.

It also uses the short-lived `sessionStorage` trick suggested in https://github.com/sveltejs/svelte.dev/pull/507#issuecomment-2429789152 to save stuff immediately before a reload. (I also tried intercepting the keypresses that would lead to a reload and preventing the default, but a) it felt a bit naughty and b) it wouldn't work on mobile.)


Rich-Harris:
will go ahead and merge because not having titles kept is annoying and this PR fixes it

PR #632, opened 2024-10-25
Type tweaks

Rich-Harris:
a few type tweaks:

- larger body copy
- shorter line width
- larger margin between paragraphs


benmccann:
I prefer the previous size on HiDPI display

The low res screen could use some tweaks though. I'd suggest making everything 10% bigger there and bumping the contrast on everything (i.e. darker text on white screen and whiter text in dark mode)

dummdidumm:
@benmccann you proposed going back to slightly darker colors - why that? Fun thing is that I did the opposite just a few days ago (basically Rich is reverting my changes here) because I felt it was a tiny bit _too_ contrasty, which might hurt prolonged reading.

benmccann:
> @benmccann you proposed going back to slightly darker colors - why that?

The font looked really blurry to me on a low res screen. I've decided the only solution is that my next laptop is definitely having hi res :laughing:

PR #634, opened 2024-10-25
docs: Add note for dimension bindings.

christophsanz:

### DIFF
--- a/apps/svelte.dev/content/tutorial/02-advanced-svelte/04-advanced-bindings/04-dimensions/index.md
+++ b/apps/svelte.dev/content/tutorial/02-advanced-svelte/04-advanced-bindings/04-dimensions/index.md
@@ -13,3 +13,7 @@ You can add `clientWidth`, `clientHeight`, `offsetWidth` and `offsetHeight` bind
 ```
 
 These bindings are readonly — changing the values of `w` and `h` won't have any effect on the element.
+
+> [!NOTE] Elements are measured using a technique similar to [this one](http://www.backalleycoder.com/2013/03/18/cross-browser-event-based-element-resize-detection/). There is some overhead involved, so it's not recommended to use this for large numbers of elements.
+>
+> `display: inline` elements cannot be measured with this approach; nor can elements that can't contain other elements (such as `<canvas>`). In these cases you will need to measure a wrapper element instead.
--- a/apps/svelte.dev/content/tutorial/02-advanced-svelte/04-advanced-bindings/04-dimensions/index.md
+++ b/apps/svelte.dev/content/tutorial/02-advanced-svelte/04-advanced-bindings/04-dimensions/index.md
@@ -14,6 +14,4 @@ You can add `clientWidth`, `clientHeight`, `offsetWidth` and `offsetHeight` bind
 
 These bindings are readonly — changing the values of `w` and `h` won't have any effect on the element.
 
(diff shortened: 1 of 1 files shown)


Rich-Harris:
Can I ask what prompted this? It's no longer true — we use `ResizeObserver`

christophsanz:
Hey!

We tried to use it on display inline and it did not work.

Changing to block, flex or grid - everything works again.

Or at least it does not work when changing the size by adding content to an inline element, or changing the style, not sure about other means of changing size.

While debugging i stumpled upon this note in the the old docs and figured that the new docs will need it as well.

So my guess is the resizeObserver has the same limitations with display inline elements as the oled approach?

Rich-Harris:
Huh! I could have sworn it worked on inline elements too — very surprising that it doesn't.

Interestingly, the `<canvas>` [works fine](https://svelte.dev/playground/hello-world?version=5.1.2#H4sIAAAAAAAAE22RzW6DMBCEX8VyUwWkCuiVAFLVS-899FB6MLCAFWIsvIQgxLvXf1VUKcf9dnZ2x95oywdQNP3eKK4SaGoBfaGCXUz1JmWkrjCgYRVT8IjXo0AQqG1opuqJSyxKUeIASBaSk4NChhAk4emP9g9pw5Uc2HrvHaMoOtq-V9RMXJny4ABtCzUGQUjygmwGlXj36ADfx4ucEZpPXAcI3HAYeYlduxv3LL4fLTInIxUXTYo9V_nmyO5QPXAd9Ys32Ofb8g9-AO96zLd-L7LYDTlLWWjlTfMsloWr_RGeaKbMjfbZ_AE-0GI2peQ1SZ5PjlSsPnfTOOu95AkAXBAbw3roD0G4IU1xmmH_2X8BusXG5uQBAAA=), so it's about the element+styles rather than just styles. Not sure how best to encapsulate that

christophsanz:
Not sure either, but some kind of small note would be nice until (or if) this can be done.

apps/svelte.dev/content/tutorial/02-advanced-svelte/04-advanced-bindings/04-dimensions/index.md
```diff
@@ -13,3 +13,7 @@ You can add `clientWidth`, `clientHeight`, `offsetWidth` and `offsetHeight` bind
 ```
 
 These bindings are readonly — changing the values of `w` and `h` won't have any effect on the element.
+
+> [!NOTE] Elements are measured using a technique similar to [this one](http://www.backalleycoder.com/2013/03/18/cross-browser-event-based-element-resize-detection/). There is some overhead involved, so it's not recommended to use this for large numbers of elements.
+>
+> `display: inline` elements cannot be measured with this approach; nor can elements that can't contain other elements (such as `<canvas>`). In these cases you will need to measure a wrapper element instead.
```
Rich-Harris, line 19:
```suggestion
> [!NOTE] `display: inline` elements do not have a width or height (except for elements with 'intrinsic' dimensions, like `<img>` and `<canvas>`), and cannot be observed with a `ResizeObserver`. You will need to change the `display` style of these elements to something else, such as `inline-block`.
```

christophsanz:
@Rich-Harris Maybe you can add this note also to the general Docs here: 
https://svelte.dev/tutorial/svelte/dimensions

Thank you!

PR #643, opened 2024-10-25
change resolution cut-off

benmccann:
This avoids switching font if you slightly zoom your browser


packages/site-kit/src/lib/styles/tokens.css
```diff
@@ -120,7 +120,8 @@
 	   the 0.01 prevents Vite from optimising it to `translate(0, 0)`, which doesn't fix anything */
 	--safari-fix: translate3d(0, 0, 0.01);
 
-	@media (resolution: 2x) {
+	// set at 1.75x so if you zoom +/-10% from common settings it won't switch your font
```
benmccann, line 123:
```suggestion
	/* set at 1.75x so if you zoom +/-10% from common settings it won't switch your font */
```

Rich-Harris (approved):

Issue #661, opened 2024-10-27
Option to exclude Tutorials from search results on svelte.dev


kwangure:
### Describe the problem

I'm frustrated when I click on a search result on `svelte.dev` and find myself looking at a tutorial. I almost never want that, even though it happens often. I just want regular docs please Mr. Search Box.

### Describe the proposed solution

A checkbox or some similar affordance that I can check and never receive tutorials in the search box results ever again! 🙈

### Importance

would make my life easier

PixelSoftwareStudios:
Even though you can just hit the dropdown arrow at the side to hide them, I agree with this. 
My proposal is adding a `docs:` prefix to the search. No checkbox, no keybind. You just type your query as normal and it will hide tutorials. I don't know if adding `tutorials:` would be as useful, but I can add that if one would like. 

I have implemented this in my PR #666

kwangure:
Searching the docs is a movement I do quite often. And when I'm looking for something very specific I'll likely do 4 - 5 searches in quick succession. I'd rather set it and forget it (i.e. localStorage or a cookie) and not have to type it every single time. I doubt anyone uses the tutorials after a few weeks of using Svelte. Svelte is good that way. :)

If this is the preferred approach though, I'd argue for no tutorials by default, and a `tutorial:` prefix to include them. I think a visual affordance is better discoverability though.

PixelSoftwareStudios:
> Searching the docs is a movement I do quite often. And when I'm looking for something very specific I'll likely do 4 - 5 searches in quick succession. I'd rather set it and forget it (i.e. localStorage or a cookie) and not have to type it every single time. I doubt anyone uses the tutorials after a few weeks of using Svelte. Svelte is good that way. :)
> 
> If this is the preferred approach though, I'd argue for no tutorials by default, and a `tutorial:` prefix to include them. I think a visual affordance is better discoverability though.

[...]

PR #666, opened 2024-10-27
Add 'docs:' search filter

PixelSoftwareStudios:
**What:** 
 A way to only search documentation results in the global search bar.
 
**Why:** 
Satisfies Issue #661

**How:** 
To test, open the search bar from anywhere, type in a phrase like "welcome" which will show both Docs and Tutorials results, then type "docs:welcome" and it will only show documentation results with that phrase.

### DIFF
--- a/packages/site-kit/src/lib/search/search.ts
+++ b/packages/site-kit/src/lib/search/search.ts
@@ -48,6 +48,11 @@ export function init(blocks: Block[]) {
  * Search for a given query in the existing index
  */
 export function search(query: string): BlockGroup[] {
+	const isDocsSearch = query.toLowerCase().startsWith('docs:');
+	if (isDocsSearch) {
+		query = query.slice(5).trimStart();
+	}
+
 	const escaped = query.replace(/[-[\]{}()*+?.,\\^$|#\s]/g, '\\$&');
 	const regex = new RegExp(`(^|\\b)${escaped}`, 'i');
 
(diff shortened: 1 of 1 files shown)


benmccann:
I agree the search is quite problematic, but this doesn't feel like the right solution to me. E.g. I'm frequently frustrated when searching for svelte docs and getting sveltekit docs. This wouldn't help at all with that.

Rich-Harris:
Agree, I'm afraid this doesn't scale — 'docs' isn't the operator, it's the _operand_. Something like this would make more sense:

- `state in:tutorial`
- `styles in:docs/svelte`
- `load in:docs/kit`

Needs to be paired with a discovery mechanism of course. Bonus points if we can offer dropdown autocompletes when you start typing `in:`.

benmccann:
I think we should also play around with other solutions. It's been very common that I'm in the Svelte docs looking for something and none of the Svelte results show up because they're buried by SvelteKit results. I don't really want to have to do more typing to make them show up. 

I agree with the user in https://github.com/sveltejs/svelte.dev/issues/661#issuecomment-2440112495. I'm an advanced user, so I never want to see the tutorial results. I think there could be a toggle or tab setting or something that is saved across searches so that I don't have to append a filter to the end of every single search that I do and can just set it a single time and have it saved.

I'd really just like to see results for the docs that I'm in by default and then have the other results in another tab. It could potentially save which one is shown by default when you do a search then.

Rich-Harris:
I'm a relatively advanced user myself and I refer back to the tutorial frequently. I think that prioritising the current section makes sense and is a quick win. I think an `in:` operator would also be useful though is less low-hanging fruit (mostly because of the discoverability requirement). I don't think we should make things convoluted with saved preferences and tabs and whatnot.

PR #668, opened 2024-10-27
styles: add transitions when switching between dark and light mode

Ocean-OS:
This adds a basic CSS transition to all elements for when dark/light mode is toggled. 

- [X] It's really useful if your PR references an issue where it is discussed ahead of time.
- [X] Prefix your PR title with `feat:`, `fix:`, `chore:`, or `docs:`.
- [X] This message body should clearly illustrate what problems it solves.

### DIFF
--- a/packages/site-kit/src/lib/styles/base.css
+++ b/packages/site-kit/src/lib/styles/base.css
@@ -28,6 +28,7 @@ body {
 }
 
 * {
+	transition: color 0.5s, background 0.5s;
 	box-sizing: inherit;
 	margin: 0;
 	padding: 0;


Rich-Harris:
Thanks. We used to have this but a) it felt gratuitous and b) you end up transitioning through some ugly intermediate states. This version in particular looks terrible when you have e.g. borders and filters and so on — you need to transition _everything_ in sync, and even then your `*` selector won't apply to pseudo elements, so you need to _also_ apply it to them (and counteract it in the places where it doesn't make sense to have transitions for whatever reason). Oh, and it means you can't really use `opacity` in lots of places, because then you get all sorts of messed up artifacts during the transition.

It _might_ be worth addressing all those issues if the transition looked good, but... it doesn't, sorry, it looks goofy. Much better to think of the toggle as a light switch rather than a sunrise/sunset.

PR #673, opened 2024-10-28
feat: add devtools updates

ignatiusmb:
Originally from https://github.com/sveltejs/svelte-devtools/pull/213
Prerequisite for https://github.com/sveltejs/svelte-devtools/issues/207

### DIFF
--- a/apps/svelte.dev/src/routes/devtools/updates.json/+server.ts
+++ b/apps/svelte.dev/src/routes/devtools/updates.json/+server.ts
@@ -1,15 +1,23 @@
+import { error, json } from '@sveltejs/kit';
+import type { ServerlessConfig } from '@sveltejs/adapter-vercel';
+
+export const config: ServerlessConfig = {
+	isr: {
+		expiration: 300
+	}
+};
+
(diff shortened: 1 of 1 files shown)


dummdidumm:
Can you add a comment to the code explaining why we need this?

ignatiusmb:
Added a comment on top of the file. I haven't really thought of other plans for `/devtools`, but I'd guess we'd want to compartmentalize this into its own route, instead of `https://svelte.dev/updates.json`.

Rich-Harris:
What happens if we're rate-limited (or the request fails for some other reason) as we're trying to prerender?

Note also that this _won't_ update when the tags update, only when the tags update _and_ we've redeployed the site, which depending on what other activity is happening could be a long delay.

We could try making it an ISR route? Otherwise I think it has to live elsewhere

Rich-Harris (approved):

Rich-Harris:
Converted it to an ISR route, checked and it works perfectly! It will revalidate every 5 minutes, so we should stay comfortably under rate limits.

PR #680, opened 2024-10-28
shrink font size when georgia is used

Rich-Harris:
@dummdidumm how's this?

### DIFF
--- a/packages/site-kit/src/lib/styles/tokens.css
+++ b/packages/site-kit/src/lib/styles/tokens.css
@@ -129,8 +129,8 @@
 
 	&.font-boring {
 		--sk-font-family-body: 'Atkinson Hyperlegible', sans-serif;
-		--sk-font-size-body: 2rem;
-		--sk-font-size-body-small: 1.8rem;
+		--sk-font-size-body: 1.9rem;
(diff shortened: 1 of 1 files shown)


dummdidumm:
yeah much better - should be done for the sans-serif font as-well.

Rich-Harris:
Why? The reason to change the size of Georgia is that it has fatter glyphs than EB Garamond. If we change the size of the sans font on low res screens it'll just mess up the ratio of body copy font size to everything else.

Unless you're saying we should change the sans font size across the board? That seemed to be an unpopular opinion before

dummdidumm:
This is how much I get to see on the old site, which was pleasant to read size-wise 
image

This is how much I get to see on the new site currently (using sans-serif, with serif it's the same)
image

It's just way too big

dummdidumm:
Reducing line-height would also help, `line-height: 1.5` is good

Rich-Harris:
Oh, interesting — didn't realise that we had 1.7 for low resolution screens — I assume the intent was to have 1.7 for Georgia and everything else, but you were seeing 1.7 on low res screens in sans mode.

Changed it to be 1.5 across the board, I think it looks fine. I nudged the sans font size down as well

PR #682, opened 2024-10-28
docs: use runes instead of old syntax in `tutorial/custom-css-transitions`

ddjerqq:
### I have updated the old syntax to use the new RUNE syntax

```js
--let visible = true;--
++let visible = $state(true);++
```


ddjerqq:
even though the code is correct, the documentation text is wrong:
# code in web editor
image

# code in tutorial text
image

Rich-Harris (approved):

Rich-Harris:
thanks!

Issue #684, opened 2024-10-28
Two links in the docs are switched


PROMCz11:
### Describe the bug

The links for the await and key block pages are switched in https://svelte.dev/docs/logic-blocks

### Reproduction

Go to https://svelte.dev/docs/logic-blocks and try to access the await page and the key page

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (8) x64 Intel(R) Core(TM) i7-8650U CPU @ 1.90GHz
    Memory: 4.07 GB / 15.92 GB
  Binaries:
    Node: 20.16.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.8.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.22621.3527
```

### Severity

annoyance

PR #688, opened 2024-10-28
fix: swap text

benmccann:
closes https://github.com/sveltejs/svelte.dev/issues/684


Rich-Harris:
made similar same change in #687, closing to avoid conflicts

Issue #65, opened 2024-07-16
Add toggle for modern AST


Rich-Harris:
see https://github.com/sveltejs/svelte/pull/12225

PR #689, opened 2024-10-28
move diagnostics logic into workspace, expose workspace.current_compiled

Rich-Harris:
DRYs some stuff out, moves some stuff out of the effect in `Editor.svelte`. Doing this ahead of tackling #65


Rich-Harris:
merging to unblock since this is just internal shuffling around

PR #695, opened 2024-10-29
Better search

Rich-Harris:
This implements a better search ranking algorithm. If you're on `/tutorial` and you search for, say, 'transitions', the top results will be from the tutorial. The same search if you're on `/docs` will prioritise docs results. But if you search for something that yields an exact title match, like `sv add`, then that will float to the top regardless of where you are.

The weights are magic numbers and there may be better ones, but these ones seem to work pretty well.

### DIFF
--- a/apps/svelte.dev/src/routes/search/+page.server.js
+++ b/apps/svelte.dev/src/routes/search/+page.server.js
@@ -11,7 +11,7 @@ export async function load({ url, fetch }) {
 
 	const query = url.searchParams.get('q') ?? '';
 
-	const results = query ? search(query) : [];
+	const results = query ? search(query, '') : [];
 
 	return {
 		query,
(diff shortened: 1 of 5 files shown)


dummdidumm:
This is much better, but it's still far from good. Typing `state` or `props` moves tutorial or SvelteKit stuff to the top, that makes no sense. In general, if you type the exact names of things, you rarely get a good result. I think we need to special-case a bunch of keywords, for example if someone types `$state` or `render` or `@render` we want the reference page for that thing to appear, but it's nowhere in sight.
image

It also doesn't help that entries are grouped by Svelte/SvelteKit etc. I don't know how the algorithm exactly determines which groups go first, but possibly you have a bunch of semi-matching stuff in one section but an exact match somewhere else but that's way below because it has less weight overall. A leading section like "top results" would help here.

dummdidumm:
Maybe the docs should have support for a new pragma that contains a list of keywords which rank this doc higher.
```
---
title: $state
keywords: $state, state
---
```

Rich-Harris:
We just need to sort the blocks _within_ the groups — previously we were just sorting the groups. So if you search for `$state`, `/docs/svelte` is the first group, but `/docs/svelte/$state` isn't the first result in the group even though it's the _reason_ `/docs/svelte` is the first group.

Now fixed

dummdidumm:
Better, but still doesn't fix
- the case where I type `state` and get everything but the `$state` page
- the case where I type `@render` and examples somehow come first, then `#snippet`, _then_ `@render`

Happy to merge this PR cause it's much better than what we have but I'm saying we're far from done, and a list of keywords might be needed.

PR #698, opened 2024-10-29
chore: spacing tweaks

dummdidumm:
- make paragraph breaks less spacey
- move `ul` closer to the preceeding paragraph which gives more visual clarity of where it belongs to
- give code blocks a bit more space at the top/bottom on larger screens
- give blockquotes more space at the top/bottom on larger screens

Combined, these changes give more visual clarity of which things belong together, which stand on its own, etc

### DIFF
--- a/packages/site-kit/src/lib/components/Text.svelte
+++ b/packages/site-kit/src/lib/components/Text.svelte
@@ -380,11 +380,11 @@
 			li::before {
 				content: '';
 				position: absolute;
-				top: 1.1rem;
+				top: calc((var(--sk-font-size-body) / var(--sk-line-height-body)) - 0.05lh);
 				left: -1.8rem;
 				background-color: var(--sk-text-4);
-				width: 0.6rem;
-				height: 0.6rem;
+				width: 0.2lh;
+				height: 0.2lh;
 				border-radius: 50%;
 				opacity: 0.7;
 			}
--- a/packages/site-kit/src/lib/components/Text.svelte
+++ b/packages/site-kit/src/lib/components/Text.svelte
@@ -119,6 +119,10 @@
 			margin: 2rem 0;
 			/* background: var(--sk-back-3); */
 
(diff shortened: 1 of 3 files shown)


packages/site-kit/src/lib/styles/base.css
```diff
@@ -45,7 +45,7 @@ p,
 ul,
 ol {
 	font: var(--sk-font-body);
-	margin: 1lh 0;
+	margin: 2rem 0;
```
Rich-Harris, line 48:
```suggestion
	margin: 0.5lh 0;
```

packages/site-kit/src/lib/components/Text.svelte
```diff
@@ -376,7 +380,7 @@
 			li::before {
 				content: '';
 				position: absolute;
-				top: 1.43rem;
+				top: 1.1rem;
```
Rich-Harris, line 383: this looks bad on my screen. we will need to find some approach that works across different configurations
 
image

dummdidumm: pushed something, let me know if that works for you

dummdidumm: saw you pushed something, how does that differ to the line-height dependent solution for you? can't we tweak the calc a bit? 
FWIW I'm fine with your solution, it's already much better than what we currently have

Rich-Harris (approved):

packages/site-kit/src/lib/components/Text.svelte
Rich-Harris: It didn't work across the board for me, and I figured if everything is the same units we don't need a `calc`. Obviously `em` and `lh` are fungible, but the way I'm thinking of it is this: `lh` for spacing in typography, `em` for things that are tied to the size of the type itself (like icons), `rem` for layout. We're not being totally disciplined but I think that's a sensible way to think about things

Issue #517, opened 2024-10-20
[omnisite docs] Contrast and font - terrible for my eyesight


lukaszpolowczyk:
### Describe the problem

"My astigmatism cannot be corrected with glasses, which results in my vision looking like this (while wearing glasses):
image
![obraz](https://github.com/user-attachments/assets/4954ee38-94be-4294-a258-1bffa5f514d6)
(double and shifted image. I see this way even when looking with one eye, it's not a squint, but rather astigmatism)

It’s worst with strong contrast, deep blacks, and bright text.

But the font is equally important. The more elaborate the font, the worse it gets.

Not only is it hard for me to read, but I actually feel discomfort and want to leave the site that has such contrast and font settings.

Light side mode is just... too bright. I don't want to use light mode anywhere, it's an unwanted compulsion in some places, but I avoid light mode.

It's disappointing because the previous version of the site had contrast and fonts that were very comfortable for me."

### Describe the proposed solution

Many websites are very well designed for my visual impairment (consciously or unconsciously), e.g.:

This Github style
![obraz](https://github.com/user-attachments/assets/bb1bf972-6773-4f25-9845-eaa79e35ef94)

ChatGPT
![obraz](https://github.com/user-attachments/assets/dc052c09-c345-45f5-9d04-d015adc78caf)

YouTube
![obraz](https://github.com/user-attachments/assets/cfde164d-7a0b-4e73-b843-02276bc76a90)

This Twitter style
![obraz](https://github.com/user-attachments/assets/c7be3bd3-9231-4fd3-9f75-b1365c64c275)

Facebook
![obraz](https://github.com/user-attachments/assets/3c312a41-d411-4afb-b206-c05a3b3e891a)

Old Svelte documentation
![obraz](https://github.com/user-attachments/assets/69dd6a21-2b02-4b3e-9fe4-fb588e09719e)

And many more.

### Importance

would make my life easier

saadeghi:
I was excited for the launch but today I couldn't even finish reading the Svelte 5 announcement blog post. This font is not a good decision.

Whateverinc:
This is my recipe: 

```

    /*change the p, ol, and ul font family to the system font or Verdana*/
    :root {
        --sk-font-family-body: system-ui, 'Verdana';
    }

    /*make the small code blocks bigger and give them a white color*/
    .text.svelte-zj4hfh code:not(pre *) {
        color: white;
        font-size: 11pt;
    }

[...]

PR #699, opened 2024-10-29
chore: tweak dark mode backgrounds

dummdidumm:
- make the darkest one a bit lighter (else it's just too dark)
- turn down background saturation to give the whole thing a bit of a warmer feeling (the blue makes it overly cold)
- remove blue from text, makes it brighter yet more pleasing to the eye

### DIFF
--- a/packages/site-kit/src/lib/styles/tokens.css
+++ b/packages/site-kit/src/lib/styles/tokens.css
@@ -151,10 +151,10 @@
 		--sk-back-6: hsl(var(--sk-back-hue), 5%, 32%);
 		--sk-back-translucent: hsla(0, 0%, 100%, 0.1);
 		--sk-theme-1-hsl: 12, 94%, 62%;
-		--sk-text-1: hsl(var(--sk-back-hue), 5%, 90%);
-		--sk-text-2: hsl(var(--sk-back-hue), 5%, 80%);
-		--sk-text-3: hsl(var(--sk-back-hue), 5%, 65%);
-		--sk-text-4: hsl(var(--sk-back-hue), 5%, 45%);
+		--sk-text-1: hsl(0, 0%, 90%);
+		--sk-text-2: hsl(0, 0%, 80%);
+		--sk-text-3: hsl(0, 0%, 65%);
+		--sk-text-4: hsl(0, 0%, 45%);
 		--sk-text-translucent: hsla(0, 0%, 100%, 0.9);
 		--sk-scrollbar: rgba(255, 255, 255, 0.3);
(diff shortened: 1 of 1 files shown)


Rich-Harris:
It doesn't feel warmer to me, just greyer. Admittedly it's hard to judge a dark theme when I'm near a window

benmccann:
I actually turned off all my lights and drew my blinds while testing this 😆

dummdidumm:
You're doing this wrong, people primarily use dark mode during the day, so it's a _feature_ of this whole thing being lighter.

> most sites are pretty dark in dark mode. e.g. I just checked https://vite.dev/guide/ and the background is pretty dark there

This PR has a _darker_ background that your example. Just proves my point that the current dark is too dark/cold. We can make the light text a notch lighter if that's what you're saying.

Rich-Harris:
> Blue == cold, less blue == warmer

In the colour theory sense, yes. I'm talking about how it _feels_. We associate bright blue skies with warmth, for example! In this case the deep blue feels to me like a luxurious velvet robe whereas the grey feels impersonal and sci-fi-ish. I don't actually think either is quite right — I can't tell if it needs to be something in between or something off to one side.

> people primarily use dark mode during the day

People baffle me.

> I just checked https://vite.dev/guide/ and the background is pretty dark

Worth noting that the situation is reversed over there — they have a darker sidebar and a lighter main content area, whereas we have the opposite.

Rich-Harris:
300 is purple and looks absolutely mad:

image

The Svelte orange hue is 12, and with 10% saturation at various lightnesses it looks... not great honestly:

image

A bluish tint works partly because it's very common and feels night-time-ish, but partly because it _contrasts_ with the orange rather than using it.

Is there an issue where people have complained about the background colour? I remember when we first introduced it (replacing the greys we'd had prior to that point) and everyone on the team was like 'oh my god this looks so much better'. Again, not saying it's perfect now — it's definitely not — but this feels like a step back towards where we were prior to that.

dummdidumm:
> Is there an issue where people have complained about the background colour?

I myself find the dark mode off-putting, it's straining on my eyes because the background is so dark, yet the text has this tint of blue and so even if it has high contrast on paper it still feels washed out.

Also these people: https://github.com/sveltejs/svelte.dev/issues/517#issuecomment-2442951290

This PR tries to find a sweet spot between having that aesthetic blue and making the background lighter to make it more pleasant on the eyes.

benmccann:
Vite's main content is lighter, but the page is on average darker because their left nav is so much darker than ours. I just tried this PR with changing the nav to a much darker color and it looks much better to me. What do you think about trying that out?

Rich-Harris:
merged #765, closing

Issue #541, opened 2024-10-22
Experiment with shorter line width for body copy, and larger paragraph margins


Rich-Harris:
our line width _is_ a bit on the large side, we could maybe nudge it down a smidge

https://x.com/hovhaDovah/status/1848814886534459719

AleksandrHovhannisyan:
Forgot to mention this, but iirc I used a line width of ~640px in my screenshot (for reference). That was just a first pass, though; it doesn't have to be that tight.

Rich-Harris:
closed via #632

PR #704, opened 2024-10-29
reduce paragraph margin

Rich-Harris:
We previously bumped the paragraph margin to `1lh` (#541), but having sat with it a while I think that was a mistake. We've had feedback — which I agree with — that the relationship between headings and copy is unclear. It also results in lower-than-ideal information density.

While `1lh` is considered something of a 'rule' in typography, I do think this works better in basically all cases (old on the left, new on the right):

image

image

image

image

image

### DIFF
--- a/packages/site-kit/src/lib/styles/base.css
+++ b/packages/site-kit/src/lib/styles/base.css
@@ -45,7 +45,7 @@ p,
 ul,
 ol {
 	font: var(--sk-font-body);
-	margin: 1lh 0;
+	margin: 0.5lh 0;
 
 	&:first-child {
 		margin-top: 0;


packages/site-kit/src/lib/styles/base.css
```diff
@@ -45,7 +45,7 @@ p,
 ul,
 ol {
 	font: var(--sk-font-body);
-	margin: 1lh 0;
+	margin: 0.5lh 0;
```
benmccann, line 48: I agree that `1lh` seems like a lot and we could do less. it does feel a little close together now, but not by much. what do you think about giving it just a tad more at `0.6lh`?
```suggestion
	margin: 0.6lh 0;
```

benmccann:
I think we could also reduce the margin between headings. E.g. the `margin-top` on `h2` could go from `7rem` to maybe `5rem`

After that, we could potentially use some of the reclaimed vertical space to tweak the `line-height` on small screens where Georgia is used, which currently feels quite scrunched. It feels a bit more comfortable to me bumping it from `1.5` to `1.65`

packages/site-kit/src/lib/styles/base.css
Rich-Harris: Typography people like to talk about 'vertical rhythm' — that's where the `1lh` came from in the first place. We're breaking that rule in a lot of places (we'd need to build the whole thing from scratch rather than merely changing the old site if we wanted to be really serious about it) but I do think that 'things should be multiples of each other' is sage advice that we should adhere to:

image


With `0.6lh`, everything becomes... bumpy. Just very slightly off:

image

dummdidumm:
Have you seen #698? Takes a more complete approach to this; there's a few other places where spacing tweaks are sorely needed

Rich-Harris:
I'm looking at it now, yeah. I think there's good stuff in there but 0.5lh works better, looking at them side-by-side (and also for the reasons given above) — in general I think we should probably prefer to use `lh` units than `rem` for typography (as opposed to layout). I'll make suggestions on that PR

Rich-Harris:
closing in favour of #698

PR #707, opened 2024-10-29
chore: light background for code snippets

dummdidumm:
While working on tweaking dark mode I noticed that I find the content much easier to scan. Turns out it's because we don't give code blocks any background whatsoever, which in my view is a mistake. It makes the code too easy to mix up with the surrounding text when scrolling and searching for something.

This gives the code blocks a very very light grey background, which in my opinion achieves a sweet spot between making it easier to separate code from text while not getting in the way. (The fact that there was a commented-out `--sk-back-3` makes me believe this was actually something on the todo list or something that was contemplated to do)

### DIFF
--- a/packages/site-kit/src/lib/components/Text.svelte
+++ b/packages/site-kit/src/lib/components/Text.svelte
@@ -112,12 +112,15 @@
 
 		.code-block {
 			position: relative;
-			background: var(--sk-back-2);
+			background: var(--sk-back-3);
 			border: 1px solid var(--sk-back-5);
 			border-radius: var(--sk-border-radius);
 			overflow: hidden;
 			margin: calc(0.5 * var(--sk-line-height-body)) 0;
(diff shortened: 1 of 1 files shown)


Rich-Harris:
The background was removed for the sake of the syntax highlighting, which suffers a bit with a background (except on dark mode, where the colours pop). I think the box works but needs to be a smidge lighter, and the bar for the filename feels too prominent.

Since those colours don't exist in our palette at the moment, I suspect we need to revisit our tokens in light of this and #699 — we have a bunch of different `--sk-text-n` and `--sk-back-n` tokens but no cohesive theory about what we're supposed to use when, and the colours themselves feel a bit random (e.g. tiny gap between 1, 2, 3 then a big jump to 4). Some uses (borders, bullets) are neither `text` nor `back` but a third thing.

Rich-Harris:
merged #765, closing

PR #710, opened 2024-10-29
fix #320

### LINKED ISSUE #320: Editors should not intercept tab presses if they were focused because of a tab press
If I focus an editor with a mouse, or if I'm actively editing, then I want tab presses to control indentation. But if I'm tabbing around the page to navigate to stuff, then tab presses should _not_ be intercepted

Rich-Harris:
fixes #320 — if you click into an editor, or type while it's focused, the tab key will then indent stuff. But if you tab into it and press again, the tab key will _not_ indent stuff — instead, focus travels to the next focusable element.

This means we meet the no-keyboard-trap requirement of WCAG 2.1 https://www.w3.org/TR/WCAG21/#no-keyboard-trap

### DIFF
--- a/packages/editor/src/lib/Editor.svelte
+++ b/packages/editor/src/lib/Editor.svelte
@@ -52,14 +52,25 @@
 	}}
 />
 
+<!-- svelte-ignore a11y_no_static_element_interactions -->
 <div
 	class="container"
 	bind:this={container}
-	onfocusin={() => {
+	onpointerdown={() => {
+		workspace.enable_tab_indent();
+	}}
+	onkeydown={(e) => {
+		if (e.key !== 'Tab') {
+			workspace.enable_tab_indent();
+		}
+	}}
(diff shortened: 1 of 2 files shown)


trueadm (approved):

packages/editor/src/lib/Editor.svelte
```diff
@@ -52,14 +52,25 @@
...
 <div
 	class="container"
 	bind:this={container}
-	onfocusin={() => {
+	onpointerdown={() => {
+		workspace.enable_tab_indent();
+	}}
+	onkeydown={(e) => {
+		if (e.key !== 'Tab') {
+			workspace.enable_tab_indent();
```
dummdidumm, line 64: how do you get out though once you're in? Should we listen to the Escape key and disable tab indent then?

dummdidumm: nevermind, this seems to work already

PR #713, opened 2024-10-30
ISR OG cards

Rich-Harris:
Turns out our OG cards are the main source of slowness in our build. This turns them into ISR routes instead. Remains to be seen if they'll work when deployed


Rich-Harris:
self-merging because these build times are killing me

PR #719, opened 2024-10-30
adjust margin/padding on `/docs`

### LINKED ISSUE #717: Horizontal Scrollbar Bug at Screen Width < 480px
https://svelte.dev/docs (only on this page)

When the viewport width is less than 480px, a horizontal scrollbar appears, which is not needed.
This issue affects the user experience on smaller devices.

Tested on firefox and chrome

Rich-Harris:
fixes #717


Rich-Harris:
just gonna do a cheeky self-merge to clear out another issue, since this ought to be uncontroversial

PR #723, opened 2024-10-30
order contributors

### LINKED ISSUE #372: Put top contributors/donors in center
I'm not personally sure we need to do this — I think there's something quite nice about saying that we value small contributions just as highly as large ones — but others may feel different

Rich-Harris:
this absolute monstrosity — which I did by hand — orders contributors and donors by number of commits/size of donation. I mostly did it as a fun CSS puzzle but if we like it we can keep it. (also if anyone has a smarter idea for how to do this I am all ears)

closes #372


trueadm (approved):
Both shocked that you got this working and also perplexed by how complicated it was!

dummdidumm:
wow 😄 Honestly I'm not sure if we want to keep that, because of the additional CSS - seems quite big.

Is there like a middle ground we can take?

Rich-Harris:
We can't know the exact impact until it's deployed, because for whatever reason the compression seems very slightly different between prod and preview deployments. But: the `.css` file that's affected by this change goes from 3.01kb transferred on `main` to 3.94kb on this branch, and with the apparently-more-effective prod compression that delta will likely go down to below 0.9kb. In other words it's basically free and we probably should make the decision based on whether we think it's a good change rather than how costly it is

dummdidumm (approved):
Approving in that case

PR #730, opened 2024-10-30
add sync-docs README

### LINKED ISSUE #573: Remove `USE_GIT`
Why do we need it?

The way I had it setup originally was that if a directory was already present, it wouldn't try to clone the repo again. That would let you symlink in any repos that you wanted to

Rich-Harris:
closes #573

### DIFF
--- a/.github/workflows/sync-docs.yml
+++ b/.github/workflows/sync-docs.yml
@@ -24,6 +24,6 @@ jobs:
         with:
           commit-message: sync ${{ github.event.client_payload.package }} docs
           title: Sync `${{ github.event.client_payload.package }}` docs
-          body: This is an automated pull request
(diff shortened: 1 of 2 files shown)


benmccann:
do we no longer have to set `USE_GIT=true` when running it?

Rich-Harris:
correct

Rich-Harris:
(I mean, if you want to pull from GitHub then you should use the `-p` or `--pull` flag, but there's no more env var)

benmccann:
Is there any reason we wouldn't make `--pull` the default? What does it do if you already have a repo and run it without any args?

Rich-Harris:
It will sync from whatever state the repo is currently in. The assumption is that if you're interacting with this stuff manually, you're working with a symlinked repo and you're probably making edits in it. In that case, you definitely don't want to pull every time you sync

trueadm (approved):

PR #741, opened 2024-10-31
Change Search Box escape key from e.code to e.key

### LINKED ISSUE #740: Change escape key Search Box behavior
Many users change caps to escape on their machine. But Search Box uses `event.code === 'Escape'` instead of `event.key === 'Escape'`. It represents physical key instead of just key. If i swap caps to escape with this command: `setxkbmap -option caps:escape` - it doesn't work. Suggestion is to change this behavior in `SearchBox.svelte` file to:
```
<svelte:window
	onkeydown={(e) => {
		if (e.key === 'k' && (navigator.platform === 'MacIntel' ? e.metaKey : e.ctrlKey)) {
			e.preventDefault();
			$search_query = '';

			if ($searching) {
				close();
			} else {
				$searching = true;
			}
		}

		if (e.key === 'Escape') {
			close();
		}
	}}
/>
```

uglyreptile:
Closes #740

### DIFF
--- a/packages/site-kit/src/lib/search/SearchBox.svelte
+++ b/packages/site-kit/src/lib/search/SearchBox.svelte
@@ -136,7 +136,7 @@ It appears when the user clicks on the `Search` component or presses the corresp
 			}
 		}
 
-		if (e.code === 'Escape') {
+		if (e.key === 'Escape') {
 			close();
 		}
 	}}


Rich-Harris (approved):

Rich-Harris:
thanks! was unaware of this distinction

PR #745, opened 2024-10-31
feat: allow version to be a `pr-[prnumber]` or `commit-[commithash]`

paoloricciuti:
So...this basically allow us to open the playground with a version which is a `pkg.pr.new` url (generated from our CI on PRs and pushes) and it will use that version of svelte.

This is working but i think there's some thing to discuss about the implementation and i wanted some feedback before (p.s. i also paired on it with @oscard0m this evening).

EDIT: i actually went ahead and implemented the solution below...so now to access a specific PR or commit you would do `https://svelte.dev/playground/hello-world?version=ref:13953`...i've also removed all the `is_pkg_pr_new` around...we can always revert that commit!

1. ~currently you should access the playground like this `https://svelte.dev/playground/hello-world?version=https://pkg.pr.new/svelte@13953` to use that version of svelte... @Conduitry was arguing that maybe we shouldn't require the whole URL so that we can detach ourselves from pkg.pr.new in the future. I think it's a fair point so maybe we can change this.~
2. ~if we stick with the whole url i made the decision to check for the specific pkg.pr.new URL to prevent people from putting urls in query param and make us fetch their url.~
3. to untar the package i've used this library [here](https://github.com/highercomve/tarparser)...it has been published something like 3 weeks ago, very early version but the code is not that complex and it seems to be doing everything alright (and it works)...it doesn't have types hence the couple of `.d.ts` that i manually generated. In theory we could remove the dependency and just copy paste the whole file in the codebase without much hassle.
4. ~I'm also passing around a lot of `is_pkg_pr_new` variables which i don't particularly like.~
5. Since i already fetch all the files from the tarball i'm using those in the rollup plugin (no need to fetch from unpkg if we already have the files)

How i think we can make this better.

1. ~we could instead of using the whole url do something like `ref:13953` as the version (i'm using ref because technically you could also use a commit hash). This will make sure that if tomorrow we want to leave pkg.pr.new we could find a different way to fetch the code for that pr/commit instead.~
2. we could use a different package to untar stuff? I didn't found one that work (it actually needs to handle `tar.gz` but if you have suggestions for a more stable package i can swap it out...we can also just copy the code.
3. ~instead of passing around `is_pkg_pr_new` everywhere we could infer it either from the version (is we change to `ref:[number]` we could just check if adhere to this format) or from the url itself (if we stick to `pkg.pr.new` i can still check that the version is an URL with that origin.~

Example:

using this link

[https://svelte-dev-git-pkg-pr-new-version-svelte.vercel.app/playground/hello-world?version=pr-13953](https://svelte-dev-git-pkg-pr-new-version-svelte.vercel.app/playground/hello-world?version=pr-13953#H4sIAAAAAAAAE7WSTW7CMBCFr2JZXSQigpZlSCJ11wN011QoxFPJqhlH9jiAoty9Mg4EWiTSBV7Nz_tm7Jd0HKst8JS_gVKa7bRRgkUgJIGIecK_pALL04-O06HxOl_gyYl6bZq5bUGRr20qC7fqtUYCJMtTntnayIaKEhljTAGxqiapkeXsyVJFEMWrsddWysHYej71vhwGKsAvEWoBSZDHrAsif3x9LhHBvMOe8pIPQJDmJWezEK5GxgA5g5dT_HGN8FdA2K3_rrm7LtAiLDzPWF1P6JPrXIAlow_R_1YNFIiS_54_pkPY37RzOdXOxzn3GIumOZItzv9oiZmQLXMW0mBN3h1v1RfZQsi2KDHbOCKNTGOtZP2dd1GcF4N2-CJ9YYHWQ5ItAjAJXV6iy4koOqUC56N70PE5s1lfSKwNbAHpTPCEE-yJp2Qc9J_9D6XrjcYqBAAA)

you can access the test of the pr 13953 (i picked that pr because it's very easy to verify that it's actually using that updated compiler)

WDYT?

Also i tried to stress test the playground a bit and it seems to work fine but i hope i'm not missing anything 😄 


Rich-Harris:
This is cool! I opened a PR on the `tarparser` repo to add types there, so that every user of the library doesn't have to duplicate the `declare module` effort: https://github.com/highercomve/tarparser/pull/1

I agree with @Conduitry that it should be `pr-nnnnn` rather than a URL:

- it's more future-proof
- it doesn't suggest to users that they _can_ use arbitrary URLs (e.g. `registry.npmjs.org` or some other source)
- it's less to type
- it doesn't involve characters that need to be URI-encoded

Similarly I don't think it should be `ref:` — PR numbers _aren't_ refs, and we shouldn't treat them as the same sort of thing as commit hashes or tags or branch names. We can always add support for refs in future if we find a need.

paoloricciuti:
> Similarly I don't think it should be `ref:` — PR numbers _aren't_ refs, and we shouldn't treat them as the same sort of thing as commit hashes or tags or branch names. We can always add support for refs in future if we find a need.

I actually already made the change to use `ref:` but I can change it to `pr-`...but since pr.pkg.new actually create one for every commit to main should we just also already add a `sha-shshsh` or `ref-272727` together with `pr-2883`?

paoloricciuti:
Ok i've implemented both `pr-83763` and `commit-kjasdha`...i've did that in two separate commits so that we can always revert commit if we don't like it.

dominikg:
`commit-xxx` is useful if you want to compare 2 different commits in the same PR, eg the initial PR contained a bug and then the author pushes a fix. If we only had `pr-xxx` then a person coming to the discussion later might never see the bug and how it plays out in the playground.

How do we handle pr.new purging versions? i don't think it keeps them around forever, so we will experience some form of link rot after PRs are closed

paoloricciuti:
> How do we handle pr.new purging versions? i don't think it keeps them around forever, so we will experience some form of link rot after PRs are closed

Mh, that's an interesting point I wonder if pkg.pr.new could serve a `@latest` which we could fall back to in case one is missing

paoloricciuti:
I was actually thinking...we should probably keep the version as the normal version and add a pr/commit query param that can override the version...so that we also have a fallback if the fetch fails. WDYT @Rich-Harris @dominikg ?

Rich-Harris:
Probably overkill — if someone is using `pr-` or `commit-` it's to see how a PR that's being actively worked on behaves. These links don't need to work forever, just while the work is being done

paoloricciuti:
> Probably overkill — if someone is using `pr-` or `commit-` it's to see how a PR that's being actively worked on behaves. These links don't need to work forever, just while the work is being done

Ok i'll just throw in a more descriptive error then...do we want to wait on the tarparser pr?

Rich-Harris:
The PR was merged but the maintainer hasn't released a new version yet, I figured we could wait until they've had a chance to do that

paoloricciuti:
> The PR was merged but the maintainer hasn't released a new version yet, I figured we could wait until they've had a chance to do that

Oh yeah absolutely...there' no rush.

paoloricciuti:
image

Did it!

Rich-Harris:
Right now the playground _only_ works if you have a `pr-` version

paoloricciuti:
> Right now the playground _only_ works if you have a `pr-` version

Uh that's weird... I'll check in a sec

paoloricciuti:
> Right now the playground _only_ works if you have a `pr-` version

Should be fixed now!

paoloricciuti:
I've also put up a PR on the svelte repo to use a custom comment that also includes a link to the playground with the version/pr already prefilled

https://github.com/sveltejs/svelte/pull/14151

Rich-Harris (approved):

Rich-Harris:
awesome!

PR #751, opened 2024-11-01
More token tweaks

Rich-Harris:
Continuing the work towards overhauling our `--sk-bg-n` tokens (https://github.com/sveltejs/svelte.dev/pull/707#issuecomment-2445580329):

- renames `--sk-border-medium` to `--sk-border`. This should be used by default for all borders unless there's a compelling reason to do something different
- renames `--sk-border-light` and `-dark` to `-highlight` and `-shadow`, to differentiate from theme terminology. These are used on the edges of buttons to give them a raised appearance, and preferably nowhere else
- removes the `--sk-bg-5` colour altogether. This was only used in one place that wasn't a border — the inactive state of `<Checkbox>` — and `--sk-bg-4` works fine
- in a couple of places we were using the highlight/shadow border colours — these are now fixed

So there are two or three very minor style adjustments, and in exchange we have a more constrained palette that will be easier to work with. The next step is to develop more consistency around when the various tokens apply, and tweak their values a bit — that can happen in a follow-up PR.


dummdidumm:
While you have the "where should X apply" in your head, we should write that into the CSS token files as comments, so that we don't regress in the future or someone new is scratching their head what token to use for a certain UI.

Rich-Harris:
Done. Will merge to unblock the next phase

PR #765, opened 2024-11-02
more css tweaks

Rich-Harris:
Tidies up a bit more CSS. This one involves actual changes:

- tweaked the light more foreground colours a bit, to add more distance between `--sk-fg-2` and `--sk-fg-3`. Also makes headings (`--sk-fg-1`) a slightly less deep black
- fixes line height for `--sk-font-body-small` (it was using the same absolute line height as `--sk-font-body`, rather than the same relative line height. This is most visible in the sidebar on the docs)
- uses `--sk-font-ui-medium` for `<figcaption>` elements. This feels appropriate and looks better IMHO
- a couple of other small fixes and tweaks
- tweaks background colours a bit. `--sk-bg-2` is no longer identical to `--sk-bg-1` in light mode, which means that e.g. code snippets have a subtle background (#707). in dark mode, there's _slightly_ less contrast, and the darker backgrounds are desaturated a tiny bit (#699)

I think this looks better but would be interested to know how others feel

### DIFF
--- a/apps/svelte.dev/src/lib/components/SecondaryNav.svelte
+++ b/apps/svelte.dev/src/lib/components/SecondaryNav.svelte
@@ -16,7 +16,7 @@
 		align-items: center;
 		justify-content: space-between;
 		padding: 0.6rem var(--sk-page-padding-side);
-		background-color: var(--sk-bg-2);
+		background-color: var(--sk-bg-1);
 		white-space: nowrap;
 		flex: 0;
 		gap: 1rem;
--- a/apps/svelte.dev/src/routes/+page.svelte
+++ b/apps/svelte.dev/src/routes/+page.svelte
@@ -46,7 +46,7 @@
 
 <style>
 	figcaption {
-		font: var(--sk-font-body-small);
+		font: var(--sk-font-ui-medium);
 	}
 
 	.home :global {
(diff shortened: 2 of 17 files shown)


benmccann:
I think this makes dark mode look nicer. Perhaps separately, I'd be curious to experiment with making the left nav darker than the main content like Vite does

In light mode, I think the right nav text doesn't have enough contrast in this PR

Rich-Harris:
tweaked both of those things, what do you reckon?

dummdidumm:
I like the changes mostly. Wondering if the left nav should have a bit more space between entries, basically revert to what it's like on main, just not by changing the line height back but by adding margins for that specifically. Semantically, the entries are not text but more like bullet points, and they have more spacing between them, too

benmccann:
Yeah, this look really nice

I was also wondering about the spacing in the left and right navs. It's not too bad either way and I'm not sure if I have a clear preference, but I definitely noticed that the items were closer together and think I probably preferred it with a bit more spacing

Rich-Harris:
I wrote this recently in response to a similar proposal about the sidebar, before I messed it up by using the main body line height instead of the `body-small` line height: https://github.com/sveltejs/svelte.dev/pull/458#issuecomment-2423430916

Normally the reason to have spacing between list items is so that it's clear that the first two lines are part of a single item, then the next two, then the next three, rather than having seven lines that are only differentiable via the placement of the bullet points:

image

That doesn't really apply when there's a single line for everything. If the feeling that we should bend the line-height rules is universal then it's easy enough to do — personally I think it's a little more consistent this way and I prefer the greater information density

PR #791, opened 2024-11-04
Fix broken link in latest blog

Devr-pro:
Fix broken link of $derived


Rich-Harris (changes requested):
Thank you. I'm going to hold off on merging this for now, because this broken link should already be causing deployments to fail — the fact that it isn't is something that needs to be investigated

Rich-Harris:
merged #804, so I'll close this. Investigating this led to a discovery that we're missing protection against many other broken links, so it was very much appreciated!

PR #817, opened 2024-11-06
feat: Add support for iOS safe area

WarningImHack3r:
Add support for iOS' safe area environment variables to prevent the navbar from being hidden behind the iPhone's home bar.

It works using the [official layout guide](https://webkit.org/blog/7929/designing-websites-for-iphone-x/), which was originally made after the introduction of the iPhone X with iOS 11.  
It exposes the `env()` CSS function with 4 variables (`safe-area-inset-{top,left,right,bottom}`), requiring `viewport-fit=cover` in the `viewport` meta tag to get rid of natural viewport margins.

These environment variables are smartly populated by the system based on the current Safari layout, giving them a value granularly as needed.

> [!NOTE]
> You might note my PR doesn't include the outdated `constant()` function, the previous name of `env()` between iOS 11.0 and 11.2 beta versions: the already low range of concerned versions combined with the ancient age of this major iOS version isn’t worth the effort, in my opinion.
> It can, however, be easily added at will with `@supports` queries or CSS rules fallbacks after bypassing `calc()` calls.

On desktop, the `env()` function is ([supposedly](https://stackoverflow.com/questions/47302707/css-safe-area-attributes-doesnt-work-on-iphone-x#:~:text=Firstly%2C%20safe%2Darea%2Dinset%2Ddir%20is%20undefined%20on%20Chrome%20and%20Safari%20on%20mac)) undefined, invalidating the CSS lines using it, leading to no repercussions on desktop browsers or other mobile browsers. I've tested it on Arc / Chrome desktop in responsive mode and noticed no impact.

## Demo

| before | after |
| :---: | :---: |
| **Before** _(old Safari layout/in-app browser)_ | **After** _(old Safari layout/in-app browser)_ |

| after old UI regular | after new UI |
| :---: | :---: |
| **After** _(old Safari layout/in-app browser - "unscrolled" state)_ | **After** _(new Safari layout - not impacted)_ |

---

### DIFF
--- a/packages/site-kit/src/app.html
+++ b/packages/site-kit/src/app.html
@@ -3,7 +3,7 @@
 	<head>
 		<meta charset="utf-8" />
 		<link rel="icon" href="/favicon.png" />
-		<meta name="viewport" content="width=device-width, initial-scale=1" />
+		<meta name="viewport" content="width=device-width, initial-scale=1, viewport-fit=cover" />
 
 		%sveltekit.head%
 	</head>
--- a/packages/site-kit/src/lib/nav/Nav.svelte
+++ b/packages/site-kit/src/lib/nav/Nav.svelte
@@ -311,6 +311,8 @@ Top navigation bar for the application. It provides a slot for the left side, th
 		nav {
 			top: unset;
 			bottom: 0;
+			padding-bottom: env(safe-area-inset-bottom);
+			height: calc(var(--sk-nav-height) + env(safe-area-inset-bottom));
 		}
 
 		.menu {
--- a/packages/site-kit/src/lib/nav/Nav.svelte
+++ b/packages/site-kit/src/lib/nav/Nav.svelte
@@ -311,8 +311,11 @@ Top navigation bar for the application. It provides a slot for the left side, th
 		nav {
 			top: unset;
 			bottom: 0;
-			padding-bottom: env(safe-area-inset-bottom);
-			height: calc(var(--sk-nav-height) + env(safe-area-inset-bottom));
+
+			&.visible, &:focus-within {
+				padding-bottom: env(safe-area-inset-bottom);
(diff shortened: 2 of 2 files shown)


WarningImHack3r:
Fixed thanks to [this forum answer](https://community.weweb.io/t/pwa-safe-area-using-env-safe-area-inset-bottom/17206/3) (for reference)

| current and pwa |
| :---: |
| **Left:** unchanged iOS 18 inside Safari / **Right:** iOS 26.2 PWA |

And while we're at it, full iOS 26 showcase (unaffected by this last patch):
| CleanShot 2026-02-26 at 23 40 31@2x | CleanShot 2026-02-26 at 23 40 46@2x |
| :---: | :---: |
| Safari 26 (large) | Safari 26 (compact) |

WarningImHack3r:
I unfortunately had to revert support for PWAs. `viewport-fit=cover` completely removes the (side) safe areas when browsing in landscape mode on a mobile device, which induces a lot of CSS issues due to absolute positioning and a now wider screen for media queries if we try to just add side padding to the `body`.
This would be a welcome change, but it is not in the scope of this PR and would require a lot more work than what I planned to do more than a year ago.

The added benefit of supporting PWAs, and more specifically only for the bottom side for the navbar on mobile, is niche compared to the amount of work it requires.

teemingc:
Any idea how to fix the gap between the navbar and the bottom? Tested on the in-app browser for ios26. This gap is an existing issue on the current svelte site too
image

WarningImHack3r:
> Any idea how to fix the gap between the navbar and the bottom? Tested on the in-app browser for ios26.

I meeean, it's in the Twitter app, no clue how they implemented their browser, nor do I have a way to debug it
I can't reproduce it in the actual native in-app browser:
In-app svelte.dev from bluesky (native iOS in-app browser)

PR #830, opened 2024-11-12
fix: mobile nav links to SvelteKit tutorials

### LINKED ISSUE #829: Mobile menu - tutorial > sveltekit links are broken
https://svelte.dev/tutorial/introducing-sveltekit

vs

https://svelte.dev/tutorial/kit/introducing-sveltekit

Sveltekit links seem to be missing the /kit...svelte tutorial links are working fine.

Conduitry:

Fixes #829. Pointing the link to the right place seemed like the correct way to address this, not fixing the redirects.

This solution feels really ugly, but I don't know how else to deal with this. The metadata that's exposed from the pages seems like it would have to change if we don't want to be looking for substrings in slugs like this. I do note that `apps/svelte.dev/src/routes/tutorial/[...slug]/content.server.ts` is already doing something similar with looking at the slug strings.

### DIFF
--- a/apps/svelte.dev/src/routes/+layout.server.ts
+++ b/apps/svelte.dev/src/routes/+layout.server.ts
@@ -28,7 +28,11 @@ const nav_links: NavigationLink[] = [
 				title: section.metadata.title,
 				sections: section.children.map((page) => ({
 					title: page.metadata.title,
-					path: '/tutorial/' + page.slug.split('/').pop()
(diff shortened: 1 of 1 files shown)


dummdidumm (approved):
yeah the "get path from slug" logic is pretty ad hoc right now, this needs an overhaul in the long run, for now let's get that quick fix in

PR #915, opened 2024-12-02
fix: add polyfills / avoid using `structuredClone`

### LINKED ISSUE #911: Svelte Playground doesn't work in Safari 17 or lower
The Svelte Playground doesn't work in Safari except for Safari 18. This is a pretty big problem as I rely on the Playground to test components and their CSS in older browsers (using browserstack.com).

This is Safari 12:

Screenshot 2024-12-02 at 14 42 53

Safari 13 and 14:

Screenshot 2024-12-02 at 14 40 37

Safari 15 and 16 and 17: it looks good but it never stops loading the compiler:

Screenshot 2024-12-02 at 14 42 07

Only in Safari 18 does it work as expected:

Screenshot 2024-12-02 at 14 46 50

dummdidumm:
- polyfills for `at` and `withResolvers`
- avoid using `structuredClone`, we can use `JSON.parse(JSON.stringify(..))` instead at that position

makes the site work in more browsers

closes #911 assuming we're fine with not supporting Safari 13 and lower (or similarly old browsers; i.e. earlier than ~2020)

### DIFF
--- a/apps/svelte.dev/src/hooks.client.js
+++ b/apps/svelte.dev/src/hooks.client.js
@@ -1,0 +1,1 @@
+import '@sveltejs/site-kit/polyfills';
--- a/apps/svelte.dev/src/routes/(authed)/playground/[id]/+page.svelte
+++ b/apps/svelte.dev/src/routes/(authed)/playground/[id]/+page.svelte
@@ -69,8 +69,8 @@
 
 		if (!hash && !saved) {
 			repl?.set({
-				// TODO make this munging unnecessary
-				files: structuredClone(data.gist.components).map(munge)
+				// TODO make this munging unnecessary (using JSON instead of structuredClone for better browser compat)
+				files: JSON.parse(JSON.stringify(data.gist.components)).map(munge)
 			});
 
 			modified = false;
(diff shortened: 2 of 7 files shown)


Rich-Harris:
I don't want to touch core-js with a bargepole. This is the effect of those two imports on the bundle size:

- before: 184.60 kB / 105.31 kB transferred
- after: 208.06 kB / 118.35 kB transferred

Let's find another approach, whether that's writing our own polyfill or avoiding those features

dummdidumm:
wow, that's bad - I created our own polyfills instead

Rich-Harris (approved):

kevinrenskers:
Can you let me know when this is deployed to the site, so that I can test it for you?

kevinrenskers:
It's definitely still broken.

Safari 14:

Screenshot 2024-12-03 at 18 55 10

Safari 15 / 16 / 17:

Screenshot 2024-12-03 at 18 56 10

Rich-Harris:
Well that's odd. I only have 17.6 installed here but it works fine. Both places that use `Promise.withResolvers` are polyfilled. Any chance there's some fucked up caching going on?

kevinrenskers:
`Promise.withResolvers` was added in Safari 17.4, so it makes sense that it works for you in 17.6.

I did a hard refresh, so I don't know? CloudFlare maybe?

simonljus:
The Polyfill for Array.prototype.at does not handle `NaN` and floating point numbers the same way as specified in the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/at#index).

PR #919, opened 2024-12-03
docs: update sample code to replace deprecated <svelte:component>

nanggo:
### What
This PR updates the sample code in the SvelteKit tutorial for universal load functions. Specifically, the `<svelte:component>` tag, which is deprecated in `runes` mode, is replaced with a recommended alternative.

### Why
Using `<svelte:component>` in the tutorial causes the following warning:

```
svelte:component is deprecated in runes mode — components are dynamic by default (svelte_component_deprecated)
```

The editor's solution already uses the updated code, but the sample code in the tutorial did not reflect this change.

### DIFF
--- a/apps/svelte.dev/content/docs/svelte/02-runes/02-$state.md
+++ b/apps/svelte.dev/content/docs/svelte/02-runes/02-$state.md
@@ -36,12 +36,7 @@ let todos = $state([
 ...modifying an individual todo's property will trigger updates to anything in your UI that depends on that specific property:
 
 ```js
-// @filename: ambient.d.ts
-declare global {
-	const todos: Array<{ done: boolean, text: string }>
-}
-
-// @filename: index.js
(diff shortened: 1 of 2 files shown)


apps/svelte.dev/content/docs/svelte/02-runes/02-$state.md
dummdidumm, line 1: these changes seem unrelated, could you revert them? (if those changes should be made, they need to happen in the Svelte repo)

nanggo: I'm sorry about that. Should I just revert or need to force push at my fork repository?
I think something wrong because 06e0046b9a6f6a41bf5343e2a7e525c5a1d5b677 commits in.

dummdidumm (approved):
thank you!

Issue #837, opened 2023-01-13
Full text in RSS feed


baukevanderlaan:
I'm grateful for the Svelte blog RSS feed, and would be even happier if the feed would serve the full text instead of only the summary. Anecdotally, a quick look in my RSS reader shows me that most blogs show the full text directly. Is this up for consideration? I'd gladly work on a PR and thought about the following options:

1. Show the full text in `https://svelte.dev/blog/rss.xml`, replacing the current style of showing only summaries.
2. Provide an additional feed accessible under a unique url: `https://svelte.dev/blog/rss-full.xml`
3. Provide an additional feed accessible with query params: `https://svelte.dev/blog/rss.xml?mode=full`

Rich-Harris:
Related (but outdated): https://github.com/sveltejs/svelte/pull/9085

PR #934, opened 2024-12-05
feat: Render full posts on RSS feed

martgnz:
Renders the full post in the RSS feed for the svelte blog. With this change the feed will be friendlier for those with unstable connections (e.g. subway) since you can download posts whenever you have a connection rather than having to navigate to an external URL each time. It's a nicer experience since you don't have to move away from your reader.

Closes https://github.com/sveltejs/svelte.dev/issues/837.

Tested on NetNewswire from localhost feed and it works well.

## before

imagen

## after

imagen

### DIFF
--- a/apps/svelte.dev/src/routes/blog/rss.xml/+server.js
+++ b/apps/svelte.dev/src/routes/blog/rss.xml/+server.js
@@ -35,7 +35,7 @@ const get_rss = async (posts) => {
 		<item>
 			<title>${escapeHTML(post.metadata.title)}</title>
 			<link>https://svelte.dev/${post.slug}</link>
-		  <author>${escapeHTML(post.metadata.author)}</author>
+			<author>${escapeHTML(post.metadata.author)}</author>
 			<description>${escapeHTML(await render_content(post.file, post.body))}</description>
 			<pubDate>${formatPubdate(/** @type {string} */ (post.file.split('/').pop()).slice(0, 10))}</pubDate>
 		</item>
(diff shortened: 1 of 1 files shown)


apps/svelte.dev/src/routes/blog/rss.xml/+server.js
```diff
@@ -25,8 +26,24 @@ function escapeHTML(html) {
...
+const get_rss = async (posts) => {
+	const renderedPosts = await Promise.all(
+		posts
+			.filter((post) => !post.metadata.draft)
+			.map(
+				async (post) => `
+		<item>
+			<title>${escapeHTML(post.metadata.title)}</title>
+			<link>https://svelte.dev/${post.slug}</link>
+			<author>${escapeHTML(post.metadata.author)}</author>
```
martgnz, line 38: added author info too

Rich-Harris (approved):

Rich-Harris:
fantastic, thank you!

Rich-Harris:
getting this rather cryptic error. not sure what could cause it 

image

martgnz:
my guess is that the render function is not happy to get called under that endpoint for some reason (maybe due to path mismatches?). on that function i saw lots of tweaks to make code snippets work on other parts of the site.

there's another markdown renderer on the project with less moving parts maybe could switch to that?

Rich-Harris:
don't have time to properly diagnose it but it turns out that prerendering it (which we should have been doing anyway) works around whatever the issue is #943

martgnz:
thanks so much Rich!

PR #936, opened 2024-12-05
fix: typo in tutorial

sahilrajput03:
Fixes the typo in tutorial.


Conduitry:
What's there is correct. It's saying that CSR can be configured similarly to how SSR is configured.

PR #937, opened 2024-12-05
fix deprecated note styles

Rich-Harris:
this fixes deprecated note styles (currently they use the quote mark icon instead of the lightbulb), and also adds a callout for any types with a `@since` annotation, which will become useful when we merge https://github.com/sveltejs/svelte/pull/14422


Rich-Harris:
(the actual content of the deprecation notices kinda sucks, those need to be fixed at the source)

PR #943, opened 2024-12-05
prerender RSS feed

### LINKED ISSUE #942: RSS feed is broken
just making sure we don't lose track of https://github.com/sveltejs/svelte.dev/pull/934#issuecomment-2521063143

Rich-Harris:
seeing if this fixes #942


Rich-Harris:
seems to do the trick

PR #948, opened 2024-12-06
Fix: grammar fix in index.md

Vanceagher:
"There's a" becomes "There are a"

Since "a couple of things" is plural, "are" is correct rather than "is." This can be hidden by the contraction, and is common in speech, but in text the error stands out.


Rich-Harris (approved):

Rich-Harris:
As a Brit, the 'correct' version seems strange in the conversational text of a tutorial, but the non-Brits among the maintainers prefer it

PR #966, opened 2024-12-09
add banner discouraging people from editing generated files

Rich-Harris:
we already have the 'edit this page' links pointing to the right repo, and a notice in this PR template discouraging edits here, but apparently that's not enough (https://github.com/sveltejs/svelte/pull/14613#issuecomment-2526183274) so this PR adds a notice to all the generated files


Rich-Harris:
gonna merge this so it doesn't keep getting stale

PR #979, opened 2024-12-10
feat: llms.txt

khromov:
Adds the following endpoints:

Listing:
https://svelte-dev-git-llms-txt-svelte.vercel.app/llms.txt

Dedicated merged docs for Svelte/Kit/CLI:
https://svelte-dev-git-llms-txt-svelte.vercel.app/llms-small.txt
https://svelte-dev-git-llms-txt-svelte.vercel.app/llms-full.txt

Package-specific docs:
https://svelte-dev-git-llms-txt-svelte.vercel.app/docs/svelte/llms.txt
https://svelte-dev-git-llms-txt-svelte.vercel.app/docs/kit/llms.txt
https://svelte-dev-git-llms-txt-svelte.vercel.app/docs/cli/llms.txt

# Suggested Advent of Svelte text

*Day XX: LLM-friendly documentation*

We've added new endpoints that make it easier for AI assistants to understand and work with Svelte documentation. These endpoints provide the documentation in a format that's optimized for Large Language Models, helping them give you more accurate and up-to-date answers about Svelte, SvelteKit, and the Svelte CLI.

#### General documentation

Abridged Svelte and SvelteKit documentation: https://svelte.dev/llms.txt
Full Svelte and SvelteKit documentation: https://svelte.dev/llms-full.txt

#### Package-specific documentation

Full Svelte documentation: https://svelte.dev/docs/svelte/llms.txt
Full SvelteKit documentation: https://svelte.dev/docs/kit/llms.txt
Full Svelte CLI documentation: https://svelte.dev/docs/cli/llms.txt


apps/svelte.dev/src/lib/server/content.ts
```diff
@@ -123,3 +131,210 @@ function create_docs() {
 export const docs = create_docs();
 
 export const examples = index.examples.children;
+
+function getSectionPriority(path: string): number {
+	if (path.includes('/docs/svelte/')) return 0;
+	if (path.includes('/docs/kit/')) return 1;
+	if (path.includes('/docs/cli/')) return 2;
+	return 3;
+}
```
khromov, line 140: This is to ensure ordering in the combined docs.

apps/svelte.dev/src/lib/server/content.ts
```diff
@@ -123,3 +131,210 @@ function create_docs() {
...
+		// Get directory paths
+		const dirA = a.split('/').slice(0, -1).join('/');
+		const dirB = b.split('/').slice(0, -1).join('/');
+
+		// If in the same directory, prioritize index.md
+		if (dirA === dirB) {
+			if (a.endsWith('index.md')) return -1;
+			if (b.endsWith('index.md')) return 1;
+			return a.localeCompare(b);
+		}
```
khromov, line 158: This is to ensure that eg `docs/svelte/01-introduction/index.md` comes before `docs/svelte/01-introduction/01-overview.md`

apps/svelte.dev/src/routes/llms.txt/+server.ts
```diff
@@ -0,0 +1,50 @@
...
+		],
+		{
+			removeLegacy: true,
+			removeNoteBlocks: true,
+			removeDetailsBlocks: true,
+			removePlaygroundLinks: true,
+			removePrettierIgnore: true,
+			normalizeWhitespace: true
+		}
+	)}`;
```
khromov, line 39: I tried to keep the default preset trimmed down in a way so that most use cases are covered but we still remove less important bits and bobs. For reference the trimmed `llms.txt` is 514KB as of today, and `llms-full.txt` (where nothing is omitted) is 843KB.

apps/svelte.dev/src/routes/llms.txt/+server.ts
```diff
@@ -0,0 +1,50 @@
...
+		],
+		minimize: {
+			removeLegacy: true,
+			removeNoteBlocks: true,
+			removeDetailsBlocks: true,
+			removePlaygroundLinks: true,
+			removePrettierIgnore: true,
+			normalizeWhitespace: true
+		}
+	})}`;
```
khromov, line 39: I tried to keep the default preset trimmed down in a way so that most use cases are covered but we still remove less important bits and bobs. For reference the trimmed llms.txt is 514KB as of today, and llms-full.txt (where nothing is omitted) is 843KB.

dummdidumm:
Thank you! I did some code style changes, nothing else so far.

A question regarding the various `llms.txt` files: According to https://llmstxt.org/ the main `llms.txt` would provide only a short overview along with links to more details (in our case, we could link to the different packages). It would also remove the need for `llms-full.txt`.

Can you elaborate on why you chose this approach over the one recommended in https://llmstxt.org/ ?

apps/svelte.dev/src/lib/server/content.ts
```diff
@@ -124,5 +126,198 @@ function create_docs() {
...
+
+	if (settings.removePrettierIgnore) {
+		minimized = minimized
+			.split('\n')
+			.filter((line) => line.trim() !== '<!-- prettier-ignore -->')
+			.join('\n');
+	}
+
+	if (settings.normalizeWhitespace) {
+		minimized = minimized.replace(/\s+/g, ' ');
```
dummdidumm, line 225: this whitespace normalization messes up the markdown, as all newlines are removed. What was the reason to add this, in which way do llms benefit from this? Or is this just about removing as much extra content as possible? (in which case, I'm not sure if this helps / is doable, because of the markdown rules)

dominikg: if these llms count tokens and whitespace isn't counted, then i am not sure we need this at all and should rather retain the more readable format for easier human consumption and possible markdown rendering to html?

khromov: 👋 Spaces are counted as tokens, so the reason to remove whitespace is purely to reduce the token count. That's why we apply the whitespace transformation to `/llms.txt` but *not* to `/llms-full.txt` in this PR.

As for raw numbers, I ran the current llms.txt through [OpenAI tokenizer](https://platform.openai.com/tokenizer) to test the effect:

Without `normalizeWhitespace`:
Tokens **197,260**
Characters **797,617**

With `normalizeWhitespace`:
Tokens **188,595**
Characters **781,754**

The reduction for token count is ~4.4%. 

As for the invalid markdown, since LLMs don't really parse Markdown, it does not need to have a detrimental effect on performance. As for the performance, there is no real way to measure without creating a "SvelteBench" benchmark that compares minified vs unminified and test it for all major LLMs. 

From my anecdotal evidence of having used the content.json file from the v4 site for 6+ months for LLM coding (which is half HTML encoded 😅 ) I haven't found a huge performance difference vs the docs.

What *does* make a difference is whether you can fit the file into context or not. If you are using a full context approach (as shown [in my recent YouTube video](https://www.youtube.com/watch?v=tprMklFzy44)) then it is absolutely required for the text file to fit inside the LLM context, otherwise you cannot use it at all. That's why I've made a compromise with a smaller `llms.txt` and a larger `llms-full.txt` for people who have huge context or use other approaches like RAG to search the docs, where the size does not matter as much.

dummdidumm: According to https://llmstxt.org/ the main llms.txt would provide only a short overview along with links to more details (in our case, we could link to the different packages). It would also remove the need for llms-full.txt. Why do we need both then?

khromov:
@dummdidumm 

> A question regarding the various llms.txt files: According to https://llmstxt.org/ the main llms.txt would provide only a short overview along with links to more details (in our case, we could link to the different packages). 

The https://llmstxt.site/ directory helpfully shows token sizes for each site. It seems like there is a long tail of very largs llms.txt but once you get down in the weeds most of them have the markdown-like format. I will change it so that we use this as well, and I will also introduce a `llms-small.txt` for people who want to use a smaller version of the docs. 

Screenshot 2024-12-12 at 23 06 45

> It would also remove the need for llms-full.txt. 

While a `llms.txt` listing would list all the packages, no systems today can actually parse that and intelligently fetch the sub-packages automatically, and I think we want to strive for broad compatibility.

`llms-full.txt` is useful for quick inclusion, eg using something like the command `@https://svelte.dev/llms-full.txt` in Cursor which would load in the docs, rather than having you type out the docs for each sub-package manually.

> That page also doesn't mention anything related to a <SYSTEM> tag.

There is no standard for this and LLMs don't parse this in any sort of organized fashion. However Claude recommends using the "XML" approach for structured input, and it's presumed they trained their model to respond to this. We can wrap things more intelligently in tags later (eg `<system>, <docs>, <notes>`) but since it's not a real format it's something we could do down the line. So either:
1. We keep it as-is
2. We remove the "XML" tagging for now
3. We go all-in on "XML" tags and also wrap the docs in `<docs>`

Happy to do whichever you prefer. 

https://docs.anthropic.com/en/docs/build-with-claude/prompt-engineering/use-xml-tags

apps/svelte.dev/src/lib/server/content.ts
khromov: 👋 I mentioned this in another comment, I will change it to have an index.

dominikg: > 👋 Spaces are counted as tokens, so the reason to remove whitespace is purely to reduce the token count. 

if whitespace counts as tokens it means make an actual difference in interpretation. esp removing newlines could end up turning a list into a single line thats "parsed" differently.

is there a way to compare both variants? would leaving newlines in only be a reasonable compromise?

khromov: @dummdidumm Implemented the index llms.txt
https://svelte-dev-git-llms-txt-svelte.vercel.app/llms.txt

apps/svelte.dev/.env
```diff
@@ -3,3 +3,5 @@ SUPABASE_KEY=
 
 GITHUB_CLIENT_ID=
 GITHUB_CLIENT_SECRET=
+
+VERCEL_URL=
```
khromov, line 7: @dummdidumm  Don't know if there's a better way of doing this, as we need VERCEL_URL in llms.txt to generate the FQDN versions of the urls.

Rich-Harris: Couldn't we just always point to svelte.dev? If someone needs one from a preview deployment, they already know how to find it, but most people never will

apps/svelte.dev/src/lib/server/content.ts
```diff
@@ -124,5 +126,198 @@ function create_docs() {
...
-
 export const examples = index.examples.children;
+
+export const packages = Array.from(
+	new Set(
+		Object.keys(docs.topics)
+			.map((topic) => topic.split('/')[1])
+			.filter(Boolean)
+	)
+);
```
Rich-Harris, line 137: this seems very elaborate — why not just this?

```suggestion
export const packages = Object.keys(docs.topics).map((topic) => topic.split('/')[1]);
```

apps/svelte.dev/.env
khromov: Sure, fixed.

apps/svelte.dev/src/lib/server/content.ts
khromov: Updated

Rich-Harris:
I've added some information to `/docs`, along with a new `/docs/llms` page. The latter could probably use some notes on how to use these files, but I'm not qualified to write them — if anyone else feels up to it, knock yourselves out. Am imagining something along these lines, if that makes sense

```md
## Claude

lorem ipsum dolor sit amet how to use with Claude

## Cursor

lorem ipsum dolor sit amet how to use with Cursor

## v0

lorem ipsum dolor sit amet how to use with Cursor

## Bolt

lorem ipsum dolor sit amet how to use with Cursor
```

dummdidumm (approved):

Issue #690, opened 2024-10-28
env tutorial is broken


dummdidumm:
https://www.reddit.com/r/sveltejs/comments/1geblhx/is_the_tutorial_broken_for_anybody_else_env_file/

PR #987, opened 2024-12-11
fix: include .env files in tutorial assets

jpaquim:
Should address the broken tutorials around `.env` handling: https://github.com/sveltejs/svelte.dev/issues/690

I added just the `.env` files to the glob import, not sure if it makes to include all dot-prefixed files, and from what I understand, Vite's glob import doesn't support the `dot: true` option in `tinyglobby`


dummdidumm (approved):
thank you!

Issue #523, opened 2024-10-17
REPL: Add Devtools embed using Chii


Ocean-OS:
### Describe the problem

Svelte's REPL often lacks in features that let the dev interact with the actual result or read logs. It makes it difficult to read console logs that aren't easily stringifiable values, and impossible if you're using the REPL on a mobile device. It also seems rather clunky for the REPL to say to open Devtools to read the log. 

### Describe the proposed solution

I've noticed in [SolidJS's playground](https://playground.solidjs.com/) that they use [Chii](https://github.com/liriliri/chii) to embed Chrome's Devtools in the REPL. I think adding this feature would add a lot more functionality to the REPL. 

### Importance

nice to have

Rich-Harris:
We investigated this in #1024 but decided against it

PR #1024, opened 2024-12-16
feat: integrate chii into playground

trueadm:
Swaps out our custom console implementation for the actual Chrome dev tools using Chii.

Closes https://github.com/sveltejs/svelte.dev/issues/523


Rich-Harris:
Should I be able to edit elements and styles?

Rich-Harris:
Noticed that the errors we emit during `$inspect.trace()` [get their names blanked out somehow](http://localhost:5173/playground/hello-world?version=5.14.1#H4sIAAAAAAAAE22QwWrDMAyGX0UVgToQHHr1ksBuO_da75A6ymJwbRMr3UbIuw83G9thN_Hx6dePVvT9jVDhCzkX4D3MbgBBg2UaSqxwtI4SqsuK_BmzlwFWP1vPMcp0J8eZXftE_3ETPJPnhAqbZGYbudNesyOGJdEMLRSJeyaxZqw5Zyg4nq2Zjpls5ZP2eShoHMmwECW0HXzbhfUpkmHJc29IPFzWbIJPwZF04U2sIKV8nNqj9sSm_i3jm-nU7S9Ysyhzh-1waOrp1GnfWB8Xhqv1g7r3bqH2jwV1hxUyfTAqnhfaXrcve9BG7VYBAAA=), which feels like a little bit of a regression ([from this](https://svelte.dev/playground/hello-world?version=5.14.1#H4sIAAAAAAAAE22QwWrDMAyGX0UVgToQHHr1ksBuO_da75A6ymJwbRMr3UbIuw83G9thN_Hx6dePVvT9jVDhCzkX4D3MbgBBg2UaSqxwtI4SqsuK_BmzlwFWP1vPMcp0J8eZXftE_3ETPJPnhAqbZGYbudNesyOGJdEMLRSJeyaxZqw5Zyg4nq2Zjpls5ZP2eShoHMmwECW0HXzbhfUpkmHJc29IPFzWbIJPwZF04U2sIKV8nNqj9sSm_i3jm-nU7S9Ysyhzh-1waOrp1GnfWB8Xhqv1g7r3bqH2jwV1hxUyfTAqnhfaXrcve9BG7VYBAAA=)).

Combined with the fact that this results in a lot of extra stuff being downloaded, and that it only partially works in Firefox, and that several of the tabs don't really _do_ anything except clutter the UI, I'm forced to wonder whether this is a good change, even though I was enthusiastic about it. Thoughts?

Leonidaz:
I think you meant this link vs the localhost link [get their names blanked out somehow](https://svelte-dev-git-playground-chii-svelte.vercel.app/playground/hello-world?version=5.14.1#H4sIAAAAAAAAE22QwWrDMAyGX0UVgToQHHr1ksBuO_da75A6ymJwbRMr3UbIuw83G9thN_Hx6dePVvT9jVDhCzkX4D3MbgBBg2UaSqxwtI4SqsuK_BmzlwFWP1vPMcp0J8eZXftE_3ETPJPnhAqbZGYbudNesyOGJdEMLRSJeyaxZqw5Zyg4nq2Zjpls5ZP2eShoHMmwECW0HXzbhfUpkmHJc29IPFzWbIJPwZF04U2sIKV8nNqj9sSm_i3jm-nU7S9Ysyhzh-1waOrp1GnfWB8Xhqv1g7r3bqH2jwV1hxUyfTAqnhfaXrcve9BG7VYBAAA=)

Overall, I typically go straight to the actual browser's Dev Tools, mostly because the logging can't print objects. It would be nice to have a console that can print everything by default, especially if it doesn't take away the panes like the actual devtools.  I think it probably makes the playground seem more "professional" and cool especially for the newer folks, so I think it would definitely help svelte's adoption as long as it's not buggy 😄 

More specifically.

- if`$inspect.trace()` can't be fixed to show the whole output (it's there but somehow not showing up until expanded) it would definitely be a bummer.
  - The vertical yellowish lines to visually distinguish nesting is also missing / not visible.
  - The labels, CreatedAtError, TracedAtError, etc. are missing / not visible.
- Just having a proper console is already a big win, IMO.
- It would be definitely nice if the console was the default tab, at least initially (if there is some kind remember my choices functionality)
- The devtools pane keeps collapsing to the button on refresh and now also on the svelte.dev site.  I could've sworn that it used to remember its height, maybe I'm dreaming.  
- If the console works in Firefox, it would still be a big win just having the console.  I tried using on 133.0.3 on a mac and it seemed to be fine, everything seemed to work same as in Chrome.
- As far as its size, hard to say, it loads pretty quickly for me and it's not something I immediately go to, so it can be loaded lazily.  I typically start looking at the code and the preview before paying attention to the console.

trueadm:
We decided that there were too many drawbacks to using Chii in the playground.

PR #1049, opened 2024-12-20
Vim UI

### LINKED ISSUE #896: Vim mode - weird behavior
With tutorial when "next button" is pressed vim mode resets (only reload helps). In playground when created a new component also resets in the original and new one.

Rich-Harris:
closes #896


Rich-Harris:
gah for some insane reason the editor locks up as soon as you type something, but only in the preview deployment (not locally). really starting to hate codemirror

Rich-Harris:
> really starting to hate codemirror

(the bug was in my code, obviously)

PR #1050, opened 2024-12-20
make prerender configurable

Rich-Harris:
This PR tweaks prerender config. By default, everything is prerendered except `/e/[slug]`, `/apps`, and `/playground/[slug]`, but if `PRERENDER=false` then it will be disabled for everything except a few server routes.

This means you can do `PRERENDER=false pnpm build` and everything builds _waaaaay_ faster, which is useful for debugging production issues locally.


Rich-Harris:
huh — prerendering by default meant a whole bunch of broken links just became apparent

PR #1051, opened 2024-12-20
feat: "download app" possibility

### LINKED ISSUE #827: REPL: Add option to download files as ZIP
Sometimes when using the REPL you may want to transfer the files to somewhere else, such as your IDE or your files, and it can be tedious to have to copy and paste the contents of each file. Because of this, I think it'd be good to have an option to download the files as a ZIP file.

dummdidumm:
closes #827

Based on the Vite template. Right now you gotta to ctrl+d to download it; once the VIM UI PR is merged, it should move into the dropdown that's part of that


dummdidumm:
It seems the `svelte-template.json` that is part of the static folder isn't showing up for some reason - but no idea why. Is it because it's gitignored?

Rich-Harris:
We'll need to incorporate the styles from the playground (which, incidentally, could probably use a spruce-up...)

Rich-Harris (approved):

PR #1063, opened 2024-12-23
update class tutorial

Rich-Harris:
updates the tutorial to go with https://github.com/sveltejs/svelte/pull/14714


dummdidumm (approved):

dummdidumm:
Right now the name is still "The class directive" which isn't totally correct (if you know that there's an actual directive), but also isn't really wrong. "The class attribute"?

Rich-Harris:
Good catch. Luckily the slug was just 'classes' so no need to set up redirects

apps/svelte.dev/content/tutorial/01-svelte/07-classes-and-styles/01-classes/index.md
```diff
@@ -14,15 +14,16 @@ Like any other attribute, you can specify classes with a JavaScript attribute. H
...
 <button
-	class="card"
-	+++class:flipped={flipped}+++
+	+++class={["card", { flipped }]}+++
 	onclick={() => flipped = !flipped}
 >
 ```
 
-This directive means 'add the `flipped` class whenever `flipped` is truthy'.
+This means 'always add the `"card"` class, and add the `flipped` class whenever `flipped` is truthy'.
```
Conduitry, line 27:
```suggestion
This means 'always add the `card` class, and add the `flipped` class whenever `flipped` is truthy'.
```

Or, alternatively, put quotes around `"flipped"`.

apps/svelte.dev/content/tutorial/01-svelte/07-classes-and-styles/01-classes/index.md
```diff
@@ -14,15 +14,16 @@ Like any other attribute, you can specify classes with a JavaScript attribute. H
 
 This works as expected — if you click on the card now, it'll flip.
 
-We can make it nicer though. Adding or removing a class based on some condition is such a common pattern in UI development that Svelte includes a special directive to simplify it:
+We can make it nicer though. Adding or removing a class based on some condition is such a common pattern in UI development that Svelte allows you to pass an object or array that is converted to a string by [clsx](https://github.com/lukeed/clsx/).
```
Conduitry, line 17:
```suggestion
We can make it nicer though. Adding or removing a class based on some condition is such a common pattern in UI development that Svelte allows you to pass an object or array that is converted to a string by [clsx](https://github.com/lukeed/clsx).
```

PR #1067, opened 2024-12-24
strip origin in more places

Rich-Harris:


Rich-Harris:
will prevent a recurrence of #1064

PR #1072, opened 2024-12-28
fix: added Svelte version check to runes widget in Playground

ThePaSch:
Currently, the playground runes notice does not take into account that Runes were not a thing pre-v5 and therefore gives instructions on how to enable runes even if the playground is using a pre-v5 version (of course, following its instructions will error the compiler). This adds a check to the widget that prevents this: if we're pre-v5, tell the user that runes aren't supported at all.


Rich-Harris (approved):

Rich-Harris:
thank you!

Issue #703, opened 2024-10-29
Theme switcher proposal


theetrain:
Currently, the theme toggle indicates when light mode or dark mode is active, but not 'system' or 'auto' user preference.

![svelte.dev theme toggle](https://github.com/user-attachments/assets/e913268f-90bb-4b91-8406-eec8d8a772dd)

I propose we switch this to a 3-button switcher: light, dark, system (default). I can contribute this change if it's agreeable.

Rich-Harris:
It _will_ respect the system preference if you select the mode that matches it (i.e. if your system is in dark mode and you're somehow in light mode, switching to dark mode will then mean that if you later switch the system to light mode the site will _also_ switch to light mode). 

If you set the site mode to something _other_ than the system preference, it will persist across system preference changes.

While that's slightly convoluted when you say it out loud, in practice I think it works well. If someone switches their system to a different mode and is annoyed that svelte.dev followed suit, their annoyance will only last as long as it takes them to flip the toggle — thereafter, things will behave as they want. 

To me that makes more sense than having a more complex UI, where you either need a dropdown (which is annoying from a design perspective — the edge of the current button lines up with everything else using `var(--sk-page-padding-side)`!) or a button that cycles through three different options (which is weird by itself) and where one of those transitions _appears_ to have no effect beyond changing the icon on the button (and what's a universally understood icon for 'system preference'?). So at that point we're talking about tooltips, which aren't ideal on mobile, and more complicated code.

I'm open to being outvoted on this and I'd be curious to see successful prior art, but my gut tells me this isn't worth it.

brunnerh:
Some examples, not sure about how "successful" they are. I suspect that they fall into the "decent enough" category.

YouTube
![YouTube top menu](https://github.com/user-attachments/assets/abdc7a9c-c26c-42fe-ba96-4d420089883e)

![YouTube theme sub-menu](https://github.com/user-attachments/assets/5f1d69b6-08db-4295-87bf-cc173542a8b5)

StackOverflow
![StackOverflow preferences page](https://github.com/user-attachments/assets/3fd88241-6954-4368-a86e-c970af52a03a)

GitHub
![GitHub appearance settings](https://github.com/user-attachments/assets/d18cef8d-868b-47db-91f5-136141a39a7d "wow...")
(Kind of complex, you can choose from different themes that are used for either the light/dark settings of the system or set it one fixed theme.)

---

[...]

PR #1076, opened 2024-12-31
fix(site-kit): respect system theme preference

theetrain:
Relates to #703.

### Problem

`system` preference isn't respected when the user's system preference changes before returning to the site.

Issue reproduction:

1. Set system theme preference to 'light'
2. Ensure svelte.dev site theme is set to light (which gets stored as `system`)
3. Leave site, or open a different site
4. Set system theme preference to 'dark'
5. Return to svelte.dev
6. Observe site theme is still 'light'

This is a common occurrence for me since my system preference is 'auto', and when I return to the site in the evening (system theme is 'dark'), svelte.dev renders in light mode.

### Solution

When loading the site with `system` as the set theme, ensure the appropriate system theme gets applied at startup.

Unfortunately, there is a flash during the system theme toggle recovery. I tried this alternative approach, but it's not any faster:

```js
let query: MediaQueryList

if (browser && $theme.preference === 'system') {
  query = window.matchMedia('(prefers-color-scheme: dark)');
  setTheme(query);
}

$effect(() => {
  if ($theme.preference === 'system') {
    setTheme(query);

    return on(query, 'change', setTheme);
  }
});
```

I suppose replacing JS-based class-switching logic to CSS-based media queries or [`light-dark()`](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/light-dark) would be an enhancement for another day.


Rich-Harris:
thanks — opened #1087 which solves the FOUC issue, so i'll close this in favour of that PR

PR #1077, opened 2024-12-31
feat: change RSS feed pubDate and items order

jsafarik:
Although I believe that the current `pubDate` format in RSS feed should be correct, I found an RSS reader not being able to parse it.

Initially I wanted to report an issue for the reader (which is not open source - nothing more to be done than submitting a report), but then I checked other RSS feeds (such as [Blender](https://www.blender.org/feed/), [Jenkins](https://www.jenkins.io/changelog/rss.xml), [Godot](https://godotengine.org/rss.xml), [DigiKam](https://www.digikam.org/index.xml), [FreeCad](https://blog.freecad.org/feed/), ...) which all have a common format (including the day of the week and seconds) - I thought it might not hurt to align Svelte's feed.

Additionally, all the above mentioned RSS feeds seem to order the items from the latest one to the oldest one, which is the reason for the added `reverse()`.


apps/svelte.dev/src/routes/blog/rss.xml/+server.js
```diff
@@ -4,11 +4,13 @@ import { render_content } from '$lib/server/renderer';
...
 
 const months = ',Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec'.split(',');
+const days = 'Sun,Mon,Tue,Wed,Thu,Fri,Sat'.split(',')
 
 /** @param {string} str */
 function formatPubdate(str) {
 	const [y, m, d] = str.split('-');
-	return `${d} ${months[+m]} ${y} 12:00 +0000`;
+	const dayOfWeek = new Date(str).getDay();
+	return `${days[dayOfWeek]}, ${d} ${months[+m]} ${y} 12:00:00 +0000`;
```
Rich-Harris, line 13: This code returns a different day of the week depending on which timezone you're in (if you're in Europe it will work, but if you're behind UTC then it won't). I think it should probably just be this instead:

```suggestion
	return new Date(str).toUTCString();
```

apps/svelte.dev/src/routes/blog/rss.xml/+server.js
```diff
@@ -4,11 +4,13 @@ import { render_content } from '$lib/server/renderer';
 export const prerender = true;
 
 const months = ',Jan,Feb,Mar,Apr,May,Jun,Jul,Aug,Sep,Oct,Nov,Dec'.split(',');
+const days = 'Sun,Mon,Tue,Wed,Thu,Fri,Sat'.split(',')
 
 /** @param {string} str */
 function formatPubdate(str) {
 	const [y, m, d] = str.split('-');
```
Rich-Harris, line 11:
```suggestion
```

Rich-Harris (approved):

Rich-Harris:
thanks!

jsafarik:
good catch with the timezones, thanks for the updates and the merge!

PR #1078, opened 2024-12-31
docs: fix https://svelte.dev/tutorial/kit/other-handlers document highlight code

dlintw:
https://svelte.dev/tutorial/kit/other-handlers


Conduitry:
image

I don't know what this is intended to do, but we don't want to just randomly highlight part of this file.

Issue #1069, opened 2024-12-25
Broken tutorial due to `data` being immutable in Svelte 5


henrykrinkle01:
Today somebody in the [Discord server](https://discord.com/channels/457912077277855764/1321395828996309137) asked why the UI doesn't update with the official solution:

https://svelte.dev/tutorial/kit/post-handlers

The solution worked in Svelte 4 because the `data` props was mutable. In Svelte 5 it's no longer the case. This issue has been brought up several times:
[#12999](https://github.com/sveltejs/kit/issues/12999)
[#14536](https://github.com/sveltejs/svelte/issues/14536)

I assume that the tutorial is migrated from Svelte 4 using the migration tool which just changes `export let data` to `let { data = $bindable() } = $props()` which currently does nothing.

I offered this answer:
```js
    let { data: _data } = $props();
    let data = $state(_data);
    $effect(() => {
        data = _data;
    });
 ```
 However, is this pattern really what should be put into the official tutorial to teach new people about Svelte/Kit?
 
 To my understanding, there are 3 possible solutions:
 - Do nothing, just use `$effect`, which I only consider as a workaround.
 - Make `data = $bindable()` work, which should make `data` and `page.data` mutable for optimistic UI updates. Best solution IMO. It worked in Svelte 4 so why can't it work in Svelte 5?
 - Something else like `$state.link`.
 
So what should be done about this tutorial in particular and the underlying problem in general?

PR #1085, opened 2025-01-05
Fix: SvelteKit post-handlers tutorial

twobitfool:
[POST handlers tutorial](https://svelte.dev/tutorial/kit/post-handlers) is broken, because `data` prop is not _deeply_ reactive.

See issue #1069 for details.


apps/svelte.dev/content/tutorial/03-sveltekit/07-api-routes/02-post-handlers/index.md
```diff
@@ -74,14 +74,18 @@ We're returning a response with a [201 Created](https://http.dog/201) status and
...
 
 		input.value = '';
 	}}
 />
 ```
 
-> [!NOTE] You should only mutate `data` in such a way that you'd get the same result by reloading the page.
+> [!NOTE] You should only update `data` in such a way that you'd get the same result by reloading the page.
+
+> [!NOTE] The `data` prop is not *deeply* reactive, so you need to replace it. Mutations like `data.todos = todos` will NOT trigger a re-render.
```
Rich-Harris, line 91:
```suggestion
> [!NOTE] You should only update `data` in such a way that you'd get the same result by reloading the page. The `data` prop is not *deeply* reactive, so you need to replace it — mutations like `data.todos = todos` will not cause a re-render.
```

Rich-Harris (approved):

Rich-Harris:
thank you!

Issue #1079, opened 2024-12-31
tutorial other-handlers can not work


dlintw:
https://svelte.dev/tutorial/kit/other-handlers

even I pressed 'solve' and refresh page, it can not

- insert record
- uppate
- delete

seems after i press the 'refresh' in the iframe icon will do the 'insert' and 'delete', but it is not a reactive way. every delete require press the refresh icon.

henrykrinkle01:
This is essentially the same as #1069

PR #1086, opened 2025-01-05
Fix: SvelteKit other-handlers tutorial

twobitfool:
[Other handlers tutorial](https://svelte.dev/tutorial/kit/other-handlers) is broken, because data prop is not _deeply_ reactive.

See issues #1079 and #1069 for more details.


Rich-Harris (approved):

Rich-Harris:
🙏

PR #1108, opened 2025-01-12
docs: Fixed TODO id in animation tutorial

F2:
The tutorial on advanced animations (see [this](https://svelte.dev/tutorial/svelte/deferred-transitions) and [this](https://svelte.dev/tutorial/svelte/animations)) is using `todo.id` without ever defining `id`.

For example, in TodoList.svelte, it says:

```svelte
<li
  class={{ done: todo.done }}
  in:receive={{ key: todo.id }}
  out:send={{ key: todo.id }}
>
```

However, the todos are defined in App.svelte like this:

```svelte
const todos = $state([
  { done: false, description: 'write some docs' },
  { done: false, description: 'start writing blog post' },
  { done: true, description: 'buy some milk' },
  { done: false, description: 'mow the lawn' },
  { done: false, description: 'feed the turtle' },
  { done: false, description: 'fix some bugs' }
]);
```

This PR adds the `id` to the todos, and also uses the id as the key in the #each loop.

Incrementing the `id` for new todos could be done in many ways, but I chose the `uid++` way to be consistent with [this example](https://github.com/sveltejs/svelte.dev/blob/1118cedd6906a3e52b89391a0c123f44e78cbd54/apps/svelte.dev/content/examples/10-animations/00-animate/%2Bassets/App.svelte#L36).

### DIFF
--- a/apps/svelte.dev/content/tutorial/02-advanced-svelte/05-advanced-transitions/01-deferred-transitions/+assets/app-a/src/lib/App.svelte
+++ b/apps/svelte.dev/content/tutorial/02-advanced-svelte/05-advanced-transitions/01-deferred-transitions/+assets/app-a/src/lib/App.svelte
@@ -2,13 +2,15 @@
 	import TodoList from './TodoList.svelte';
 
 	const todos = $state([
-		{ done: false, description: 'write some docs' },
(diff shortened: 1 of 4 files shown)


eltigerchino (approved):
Thank you!

PR #1127, opened 2025-01-27
docs:Update App.svelte in Clock example according new rune syntax

pochemuto:
Current example doesn't have legacy "$:" syntax. The comment is misleading.

### DIFF
--- a/apps/svelte.dev/content/examples/12-svg/01-clock/+assets/App.svelte
+++ b/apps/svelte.dev/content/examples/12-svg/01-clock/+assets/App.svelte
@@ -4,7 +4,7 @@
 	let time = $state(new Date());
 
 	// these automatically update when `time`
-	// changes, because of the `$:` prefix
+	// changes, because of the $derived
 	let hours = $derived(time.getHours());
 	let minutes = $derived(time.getMinutes());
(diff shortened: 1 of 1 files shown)


benmccann (approved):

Conduitry:
Do we want to rework this a bit further and use https://svelte.dev/tutorial/svelte/reactive-builtins instead?

Rich-Harris:
We actually don't gain a lot by doing so in this case — the diff looks like this:

```diff
<script>
	import { onMount } from 'svelte';
+	import { SvelteDate } from 'svelte/reactivity';

-	let time = $state(new Date());
+	let time = $state(new SvelteDate());

	// these automatically update when `time`
	// changes, because of the `$:` prefix
	let hours = $derived(time.getHours());
	let minutes = $derived(time.getMinutes());
	let seconds = $derived(time.getSeconds());

	onMount(() => {
		const interval = setInterval(() => {
-			time = new Date();
+			time.setTime(Date.now());
		}, 1000);

		return () => {
			clearInterval(interval);
		};
	});
</script>
```

So I'll merge this as-is — thanks!

Rich-Harris (approved):

PR #1130, opened 2025-01-29
fix: correct redirection for examples/[...path]

### LINKED ISSUE #1129: Requests to /examples/[...] not being redirected correctly
It seems that requests to /examples/[...] are not being redirected properly.

https://svelte.dev/examples/hello-world

This might be related to #1056.

tomoam:

---

fixes #1129 .

The following need to be checked on the preview site:
- `/examples` redirects to `/playground`
- `/examples/hello-world` redirects to `/playground/hello-world`
- `/examples/thumbnails/actions.jpg` is not redirected

### DIFF
--- a/apps/svelte.dev/src/hooks.server.js
+++ b/apps/svelte.dev/src/hooks.server.js
@@ -61,6 +61,11 @@ export async function handle({ event, resolve }) {
 		redirect(307, event.url.pathname.replace('/repl', '/playground'));
 	}
 
+	// For examples
+	if (event.url.pathname.startsWith('/examples')) {
+		redirect(307, event.url.pathname.replace('/examples', '/playground'));
+	}
+
 	// Best effort to redirect from Svelte 3 tutorial to new tutorial
 	if (event.url.pathname.startsWith('/tutorial/') && event.url.pathname.split('/').length === 2) {
 		redirect(307, event.url.pathname.replace('/tutorial/', '/tutorial/svelte/'));
--- a/apps/svelte.dev/src/hooks.server.js
+++ b/apps/svelte.dev/src/hooks.server.js
@@ -55,8 +55,8 @@ export async function handle({ event, resolve }) {
 		redirect(307, destination);
 	}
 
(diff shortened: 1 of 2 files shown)


apps/svelte.dev/src/hooks.server.js
```diff
@@ -55,8 +55,8 @@ export async function handle({ event, resolve }) {
 		redirect(307, destination);
 	}
 
-	// For REPL. For some reason, the repl/+page.server.ts file is not working, so
-	// we are doing the redirect here
+	// For REPL. In SvelteKit, redirects can also be handled in +page.server.js,
+	// for example, see routes/examples/[...path]/+page.server.ts file.
```
dummdidumm, line 59: that comment was correct before, it talks about the fact that for some mysterious reason putting the reidrect into `rep/+page.server.ts` doesn't work.
```suggestion
	// For REPL. For some reason, the repl/+page.server.ts file is not working, so
	// we are doing the redirect here
```

apps/svelte.dev/src/routes/examples/[...path]/+page.server.ts
```diff
@@ -1,5 +1,7 @@
 import { redirect } from '@sveltejs/kit';
 
+export const prerender = false;
```
Comment (dummdidumm) [inline apps/svelte.dev/src/routes/examples/[...path]/+page.server.ts:3]: ah yea, that was missing, of course!

I'm wondering though if we should move it into the `handle` hook aswell and delete this file, because AFAIK all redirects happen inside `handle` right now.

Comment (tomoam) [reply apps/svelte.dev/src/routes/examples/[...path]/+page.server.ts:3]: Thanks for the suggestion. For consistency, it does seem better to consolidate the redirects into the `handle` hook.  I’ll update the PR to move the redirect logic there and remove the file accordingly.

Comment (tomoam) [reply apps/svelte.dev/src/routes/examples/[...path]/+page.server.ts]: I’ve completed the changes. Please review at your convenience.

Rich-Harris (approved):

Rich-Harris:
thanks!

PR #1132, opened 2025-01-30
docs: link to old site in welcoming directory

benmccann:
People are having trouble finding the v3/v4 docs. Hopefully this helps some


Conduitry (changes requested):
I think the typographic curly quotes were intended. For content coming from Markdown, we have a special processor to replace straight quotes with curly ones - but we need to deal with other files manually.

Conduitry:
Oh, there also already were a couple here in this file that were straight quotes. We should change those to be curly instead.

dummdidumm (approved):

PR #1154, opened 2025-02-09
tutorial: replace `onMount` with `$effect`

artemszelenov:


artemszelenov:
@benmccann hey! I thought `onMount` is legacy one after runes introduced

dummdidumm:
This is fine as-is, and also makes the intent clearer - that this only runs once on mount. `onMount` is not deprecated. Therefore closing, thank you.

PR #1159, opened 2025-02-14
fix_install_command index.md

inexistentHeap404:
npx create-svelte myapp                                                                                                                                                                                      ─╯
Need to install the following packages:
create-svelte@7.0.0
Ok to proceed? (y) y

npm warn deprecated create-svelte@7.0.0: create-svelte has been deprecated - please begin using https://www.npmjs.com/package/sv instead 'npm create svelte' has been replaced with 'npx sv create'

this was not updated in the web version too.


Conduitry:
There's no diff in this PR. What were you trying to change?

PR #1188, opened 2025-03-14
chore: update workflow permissions again

Rich-Harris:
@pngwn said this might work


pngwn:
This won't stand up in court.

PR #1204, opened 2025-03-18
fix: convert type assertions

Rich-Harris:
This converts `/** @type {Foo} */ (foo)` to `foo as Foo`, which is necessary for a PR I plan to make on `sveltejs/svelte` shortly


Rich-Harris:
Have checked a bunch of code blocks to be sure this doesn't bork anything up, and it looks good — self-merging to unblock the other mentioned PR

Issue #1183, opened 2025-03-11
Add padding-right on the floating ToC


hyunbinseo:
Shouldn't there be a padding-right on the floating ToC? Example from the [FAQ page](https://svelte.dev/docs/kit/faq).

![Image](https://github.com/user-attachments/assets/d8d1921b-9f56-4d4f-b60c-4f7d0a7ca3a9)

PR #1206, opened 2025-03-19
fix: adjust right padding in OnThisPage component to fix text overflow

Leenus533:
### Issue Reference

Fixes [#1183](https://github.com/sveltejs/svelte.dev/issues/1183)

On certain screen resolutions and specific pages, the "On This Page" navigation text appears to overflow or sit too close to the container edge. 
### Description of Changes

- Added a right padding of `1rem` to the "On This Page" navigation container.
- This ensures proper spacing and improves readability, preventing text from appearing too close to or overflowing the container edge.

### Screenshots

| Before                                                                                       | After                                                                                        |
| -------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------- |
| Before fix | After fix |


Rich-Harris:
thanks! it should actually have _more_ padding, so I added that and used `text-indent` to make it clearer when items wrap

Issue #77, opened 2024-09-09
Add `"sideEffects": false` to libraries


benmccann:
We should add `"sideEffects": false` in the `package.json` of `site-kit` and `repl` as it results in Vite producing smaller output

PR #1211, opened 2025-03-21
remove sideEffects: false

### LINKED ISSUE #1210: svelte tutorial not working on windows 8.1
Hello,

When I launch the svelte tutorial on https://svelte.dev/tutorial/svelte/welcome-to-svelte, it seems to get stuck on "loading  svelte compiler".

Image

in the console, I can see 2 error messages 

Image

so it seems the tutorial breaks because Promise.withTransformers is not available.

This could be the explanation of why it does not work on windows 8.1 because the main browsers on 8.1 are stuck on LTS versions specific to windows 8.1 like chrome 109 and firefox 115

According to https://caniuse.com/mdn-javascript_builtins_promise_withresolvers withResolvers is not available on these versions.

Would it be possible in the case that withResolvers is not available, the tutorial would use a polyfill like

```ts
if (typeof Promise.withResolvers === "undefined") {
  Promise.withResolvers = <T>() => {
    let resolve!: (value: T | PromiseLike<T>) => void;
    let reject!: (reason?: unknown) => void;
    const promise = new Promise<T>((res, rej) => {
      resolve = res;
      reject = rej;
    });
    return { promise, resolve, reject };
  };
}
```

---
After writing this, I realized that you already have a polyfill for this in the code in https://github.com/sveltejs/svelte.dev/blame/main/packages/site-kit/src/lib/polyfills/index.ts that was commited 3 months ago.

could it be that this commit has not yet reached the tutorial on https://svelte.dev/tutorial/svelte/welcome-to-svelte ?

Rich-Harris:
we added these in #181 because per #77 it results in 'smaller output'...

...because it's omitting code we need. This will close #1210, but we should see what the impact on bundle size is before committing

### DIFF
--- a/packages/repl/package.json
+++ b/packages/repl/package.json
@@ -1,7 +1,6 @@
 {
 	"name": "@sveltejs/repl",
 	"version": "0.6.0",
-	"sideEffects": false,
 	"scripts": {
 		"dev": "vite dev",
 		"build": "vite build && npm run package",
--- a/packages/site-kit/package.json
+++ b/packages/site-kit/package.json
@@ -1,7 +1,6 @@
 {
 	"name": "@sveltejs/site-kit",
 	"version": "6.0.0-next.64",
-	"sideEffects": false,
 	"description": "Styles and components for use in *.svelte.dev websites",
 	"type": "module",
 	"scripts": {


Rich-Harris:
this modestly increases bundle sizes but not to the extent that I think it should interfere with this

PR #1214, opened 2025-03-24
feat: add templating mode selector

paoloricciuti:
This allow us to play around and see the output of with https://github.com/sveltejs/svelte/pull/15538 and https://github.com/sveltejs/svelte/pull/15599 ...when we merge something like this we can change this PR to only show the selector in those specific versions.


packages/repl/src/lib/Output/Viewer.svelte
```diff
@@ -102,6 +102,7 @@
 	$: if (ready) proxy?.iframe_command('set_theme', { theme });
 
 	async function apply_bundle($bundle: Bundle | null | undefined) {
+		console.log({ $bundle });
```
Rich-Harris, line 105:
```suggestion
```

Rich-Harris:
The `html` option should be checked by default:

image

Might also be good to bump up the default height of the compiler options pane to accommodate the extra row

Ocean-OS:
Found a bug where moving from one example to the other removes the option entirely, found by moving from the "hello world" example to the "reactive assignments" example

paoloricciuti:
> Found a bug where moving from one example to the other removes the option entirely, found by moving from the "hello world" example to the "reactive assignments" example

Thanks...I also showed it with latest and that fixed the issue...if we can merge this today so i can show it on TWIS it would be great (otherwise I'll use the preview version)

Rich-Harris (approved):

PR #1218, opened 2025-03-24
Tailwind

Rich-Harris:
We can now use Tailwind in the playground — just click the toggle:

image

It probably makes sense to store the option along with the data (and in the hash, where applicable); am currently building this at a hackathon and may or may not get as far as that


Rich-Harris:
gonna YOLO-merge this for the hackathon demo, forgive me

PR #1219, opened 2025-03-24
persist tailwind option

Rich-Harris:
Follow-up to #1218 — we need to store the option in the database and in hashes. So far this does the hash (though it doesn't currently update with just the toggle, you need to edit the file as well), the database stuff is probably something we should do more carefully than under hackathon conditions


Rich-Harris:
gonna do another YOLO-merge

PR #1228, opened 2025-03-25
use module workers

Rich-Harris:
trying this to see if it results in Tailwind being loaded lazily. fingers crossed...


Rich-Harris:
I briefly wondered if we should change the build config in SvelteKit, but that _could_ be a breaking change — updating it here in the Vite config seems safest

PR #1240, opened 2025-03-31
bump shiki and twoslash

Rich-Harris:
Bumping these as part of an attempt to figure out why `/blog/rss.xml` keeps failing to build on Vercel (apparently we're indirectly using `createShikiInternalSync` somewhere rather than `createShikiInternal`, and it needs to have the languages loaded upfront? except that sometimes it works and I have no idea how to reliably reproduce it)


Rich-Harris:
looks like everything is working exactly as before, merging

PR #1244, opened 2025-04-01
switch from unpkg to jsdelivr

Rich-Harris:
This replaces unpkg.com with jsdelivr.com, which is actively maintained and seems somewhat more robust (I've been experiencing a lot of issues with unpkg recently and I'm apparently not alone).

At least in local testing, jsDelivr also seems faster. However:

1. jsDelivr doesn't resolve directory imports. This means that a file like [this](https://cdn.jsdelivr.net/npm/radix-svelte@0.9.0/dist/index.js)...

```js
export * from './components';
```

...doesn't work, because it should be `./components/index.js`. Personally I'm not too concerned about this — packages should contain valid imports.

2. More concerningly, it doesn't always successfully resolve entry points. It can resolve [`svelte`](https://cdn.jsdelivr.net/npm/svelte) but not [`svelte/legacy`](https://cdn.jsdelivr.net/npm/svelte/legacy), for example — it doesn't appear to handle `pkg.exports` correctly. Appending `/+esm` to pathnames results in a 200 response, but not one we can use (it bundles stuff in a way that would break things).

### DIFF
--- a/apps/svelte.dev/src/lib/tutorial/adapters/rollup/index.svelte.ts
+++ b/apps/svelte.dev/src/lib/tutorial/adapters/rollup/index.svelte.ts
@@ -23,7 +23,7 @@ export async function create(): Promise<Adapter> {
 	let done = false;
 
 	bundler = new Bundler({
-		packages_url: 'https://unpkg.com',
+		packages_url: 'https://cdn.jsdelivr.net/npm',
 		svelte_version: 'latest',
 		onstatus(val) {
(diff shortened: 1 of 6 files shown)


Rich-Harris:
Actually I take back point 2, we can use the same mechanism we currently use (fetching the package.json and resolving from that). So it's only really the directory imports, which we can choose to solve or not. If we choose to solve it, we can use the jsDelivr API.

I think there might be a better alternative though — PR incoming

PR #1252, opened 2025-04-03
docs: Improve explanation of 05-keyed-each-blocks in tutorial

anshll:
I personally was confused with the previous explanation as to what it meant that the other list items were updated. Explaining it that egg replaced the name attribute for the doughnut, and doughnut replaced the name attribute for the car node, helped me understand what Svelte was actually doing. Explicitly drawing out what Svelte is doing made it make more sense, but I'm open to feedback! If this doesn't fit the style or is too verbose, I'll try to figure out something else! 

### DIFF
--- a/apps/svelte.dev/content/tutorial/01-svelte/04-logic/05-keyed-each-blocks/index.md
+++ b/apps/svelte.dev/content/tutorial/01-svelte/04-logic/05-keyed-each-blocks/index.md
@@ -2,7 +2,7 @@
 title: Keyed each blocks
 ---
 
-By default, when you modify the value of an `each` block, it will add and remove DOM nodes at the _end_ of the block, and update any values that have changed. That might not be what you want.
+By default, when you modify the iterable the `each` block displays, it will add or remove DOM nodes at the _end_ of the block, and update all nodes' values to reflect the modification. That might not be what you want.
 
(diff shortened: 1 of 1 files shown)


apps/svelte.dev/content/tutorial/01-svelte/04-logic/05-keyed-each-blocks/index.md
```diff
@@ -11,6 +11,8 @@ Click the 'Remove first thing' button a few times, and notice what happens:
 1. It removes the last component.
 2. It then updates the `name` value in the remaining DOM nodes, but not the emoji.
```
Rich-Harris, line 12:
```suggestion
2. It then updates the `name` value in the remaining DOM nodes (the text node containing 'doughnut' now contains 'egg', and so on), but not the emoji.
```

Rich-Harris (approved):
thanks!

kelzenberg:
Sorry to comment on this PR, but this still puzzles me and I haven't figured out how to improve it in a new PR yet...

I get that we remove the _first_ element from the `things` array and svelte always removes the _last_ DOM node from `each` blocks. We are left with four DOM nodes, for which the `name` properties are being updated in order. Easy.

What I don't get, is, **_where_** the line is being drawn from svelte on the mentioned "fine-grained updates". Because it is certainly not revisiting this line for some reason:
```
14 |  const emoji = emojis[name];
```
Is this an object reference thing or where am I stuck here?

PR #1253, opened 2025-04-03
fix: bug where 'time' was in plural when count is equal to 0

Saterz:
While doing the course to learn Svelte I saw in the button : 'Clicked 0 times' that 'time' was in plural, and normally (except if something change in the English language) this isn't suppose to be the case.

Here is a screenshot of how it looked like before the fix :
Screenshot from 2025-04-03 13-24-41

And after the fix :
[Screencast from 2025-04-03 13-26-45.webm](https://github.com/user-attachments/assets/797e3dd4-9769-4a06-8fe8-bdaf41d9b847)


Rich-Harris:
'0 times' is correct. You use the singular when there's one of something, that's why it's called the singular

PR #1255, opened 2025-04-04
update adapter-vercel

Rich-Harris:
testing out https://github.com/sveltejs/kit/pull/13679


Rich-Harris:
this looks good, but in order to properly test it it needs to be the production deployment, so i'll merge this and open an issue to use `@latest` once we can

PR #1259, opened 2025-04-04
simplify Canvas.svelte in actions tutorial

Rich-Harris:
small thing i noticed


Hugos68:
Isn't it even simpler and more modular to use `on`?
```js
$effect(() => on(window, 'resize' () => {
    canvas.width = window.innerWidth;
    canvas.height = window.innerHeight;
}));
```

Rich-Harris:
No, because that won't result in the function being called when the component first mounts

PR #1269, opened 2025-04-07
remove shiki@1

Rich-Harris:
Am trying to diagnose some annoying build flakiness — it's something to do with shiki failing to load language modules. As a first step this removes version 1, since we use version 3 elsewhere (and the package that imports version 1 doesn't appear to use it?).

There's also some trickery around `ssr.external` that I don't understand


Rich-Harris:
> require is not defined

sigh

Rich-Harris:
Okay I think our builds are actually reliable now. Man, shiki is an absolute beast

Rich-Harris:
self-merging to unblock docs PRs

Issue #868, opened 2024-11-15
Tutorial "solve" button unresponsive and console errors if active file disappears


dslatkin:
### Describe the bug

Under some conditions, if you have an active file in a tutorial exercise that would otherwise disappear when you click "Solve", clicking solve doesn't do anything and console errors a message. If you change to a file that will persist in the solution, then the solve button works as expected.

### Reproduction

1. Go to [this playground page](https://svelte.dev/tutorial/svelte/universal-reactivity)
2. Click the `shared.js` file in the sidebar
3. Click "Solve" and observe the browser's console error
4. Click the `Counter.svelte` file in the sidebar
5. Click "Solve" and notice that it works

### Logs

```shell
utils.DjNpXng4.js:1 Uncaught Error: Invalid selection /src/lib/shared.js
    at Wk.set (Editor.BZXAGenm.js:426:10342)
    at HTMLButtonElement.toggle (25.DO2QsNE4.js:15:8330)
    at b.__click (25.DO2QsNE4.js:15:3898)
    at HTMLDivElement.N (utils.DjNpXng4.js:1:1472)
```

### System Info

```
Browser version info:

Arc	130.0.6723.117 (Official Build) (arm64) 
Revision	37382504fe66965a688029d08706ddadca7ceb46-refs/branch-heads/6723@{#1765}
OS	macOS Version 15.0 (Build 24A5264n)
JavaScript	V8 13.0.245.18
User Agent	Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/130.0.0.0 Safari/537.36
```

### Severity

annoyance

Issue #1164, opened 2025-02-21
Solve Button on tutorial website doesn't work - Topic : Rest Parameters


SikandarJODD:
### Describe the bug

Visit : https://svelte.dev/tutorial/kit/rest-params

1. Solve button doesn't work as expected 

https://github.com/user-attachments/assets/b16ef693-4736-487a-8b89-35cb347a00d0

Solution : It should update ```[path] -> [...path]```

### Reproduction

https://svelte.dev/tutorial/kit/rest-params

### Logs

```Shell

```

### System Info

```Shell
Windows
```

### Severity

annoyance

mootari:
This only happens if clicking Solve would remove the currently selected file.

Note that the issue description is incomplete, as clicking the button logs an error to the console:
> Uncaught Error: Invalid selection /src/routes/[path]/+page.svelte

A similar issue exists on https://svelte.dev/tutorial/kit/optional-params:
> Uncaught Error: Invalid selection /src/routes/+page.svelte

eltigerchino:
fixed by https://github.com/sveltejs/svelte.dev/pull/1297

PR #1297, opened 2025-04-12
fix: repl: Always select a file when calling workspace.set()

### LINKED ISSUE #1287: Tutorial Bug: playground doesn't update on transition from Universal Reactivity to Default Values
Here is the buggy state, notice that the playground still corresponds to the files/code from the previous step in the tutorial.

Image

When I refresh the page, then I get the correct playground files/code loaded.

Image

cmcaine:
This might fix #1287, #868, etc.


cmcaine:
The way to test the bug is to go to the tutorial, try selecting some files and then click forward and back and see what happens.

Navigating between most of the tutorials now seems to work correctly, but going [here](https://svelte-dev-git-fork-cmcaine-patch-1-svelte.vercel.app/tutorial/svelte/actions), selecting Canvas.svelte and then clicking the next arrow shows me a JSON file for some reason.

zqianem:
> going [here](https://svelte-dev-git-fork-cmcaine-patch-1-svelte.vercel.app/tutorial/svelte/actions), selecting Canvas.svelte and then clicking the next arrow shows me a JSON file for some reason.

The strange behavior here is avoided by #1352 (fixes the error when navigating), but this fix is still the correct one for #868 and #1164 (fixes the error when clicking "solve").

packages/repl/src/lib/Workspace.svelte.ts
```diff
@@ -425,13 +425,9 @@ export class Workspace {
...
-			if (!file) {
-				throw new Error(`Invalid selection ${selected}`);
-			}
-			this.#select(file as File);
+		const matching_file = selected && files.find((file) => is_file(file) && file.name === selected);
+		if (matching_file) {
+			this.#select(matching_file as File);
 		} else {
 			this.#select(first);
 		}
```
zqianem, line 433: If we trust the types (`selected` can't be undefined because `this.#current` is never `nullish`), this can be further simplified to:

```suggestion
		const file = files.find((file) => is_file(file) && file.name === selected) as File;
		this.#select(file ?? first);
```

and the `?` in this method's default argument can be removed.

eltigerchino (approved):
Thank you so much!

cmcaine:
Thanks for reviewing and accepting this PR!

packages/repl/src/lib/Workspace.svelte.ts
zqianem: Ah yes, you're right, I completely missed the initialization and was only looking at assignments to `#current`.

PR #1300, opened 2025-04-14
feat: add reset button to REPL

### LINKED ISSUE #1126: Svelte 5 REPL: reload button
### Describe the problem

While messing with REPLs, I often want to "reload the page" / reset the state. However, the only way to do this currently is to add or remove a meaningless space in a code.

### Describe the proposed solution

Add a button that "reloads the page" / resets the state.

### Importance

nice to have

7nik:
Closes #1126
The button just re-mounts the App, but maybe re-instantiating the entire iframe, to clean up forgotten timers etc., would be better.
The button location isn't the best one - there is no address bar, like in the tutorial or SvelteLab, so alternatives are the console head and before the buttons Fork and Save the REPL. But I find them a bit too far from the tab.

### DIFF
--- a/packages/repl/src/lib/Output/Output.svelte
+++ b/packages/repl/src/lib/Output/Output.svelte
@@ -66,6 +66,8 @@
 
 	let current = $derived(workspace.current_compiled);
 
+	let resultTab: Viewer;
+
 	// TODO this effect is a bit of a code smell
 	$effect(() => {
 		if (current?.error) {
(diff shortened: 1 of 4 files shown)


7nik:
I'm still not sure what is better: remounting app (simplier) or replacing iframe (would clean up globals, timers, intervas, etc).

dummdidumm:
Reloading the entire iframe would be better since intervals is probably the thing I'd use this most for

7nik:
hm, something went not really well.

PR #1302, opened 2025-04-14
Consolidate icons

Rich-Harris:
Been meaning to do this for a while — we can deduplicate various icon files across the repo by having a dedicated `icons` package that serves `apps/*` _and_ `packages/*`. 

By using CSS masks we can also avoid the need to have separate light and dark variants.

I'm doing this now because some upcoming work will involve adding a bunch more icons and I'd like to be able to change colours on hover; if I don't do it soon it will create a whole bunch of extra work


Rich-Harris:
So much happier with this — as well as deduplicating stuff it fixes a few rough edges around colours not being switched in dark mode

PR #1312, opened 2025-04-18
fix: improve underlining in "On this page" 

zqianem:
This ensures that the anchor that is clicked is always the one that gets underlined, even if just scrolling to that heading would underline a different anchor.

A good test page for this https://svelte.dev/docs/kit/introduction; set the browser window size to 1200x1200 and try to click on "What is SvelteKit" in the right sidebar.

### DIFF
--- a/apps/svelte.dev/src/routes/docs/[topic]/[...path]/OnThisPage.svelte
+++ b/apps/svelte.dev/src/routes/docs/[topic]/[...path]/OnThisPage.svelte
@@ -6,17 +6,28 @@
 
 	let headings: NodeListOf<HTMLHeadingElement>;
 	let current = $state('');
+	let locked = $state(false);
 
 	afterNavigate(() => {
-		current = location.hash.slice(1);
 		headings = content.querySelectorAll('h2');
-		update(); // Ensure active link is set correctly on navigation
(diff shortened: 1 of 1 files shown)


zqianem:
Did some more testing on this and the results are actually browser-dependent based on the relative timings of the `hashchange` and `scroll` events. The bad behavior is consistent on Firefox, but doesn't happen most (but not all!) of the time on Chromium.

Even so, there is still a flash of underlining on the wrong anchor before the one clicked on is underlined, and the wrong anchor will always be underlined if loading the page from a link with an anchor (e.g. [current](https://svelte.dev/docs/kit/packaging#Publishing) vs. [this PR](https://svelte-dev-git-fork-zqianem-fix-anchor-link-update-svelte.vercel.app/docs/kit/packaging#Publishing)).

Either way, this PR unifies and fixes the behavior — the anchor link that is clicked is always the one underlined.

teemingc:
Closing in favour of https://github.com/sveltejs/svelte.dev/pull/1555 which fixes the behaviour without relying on clicking

zqianem:
Hmm, I'm still running into the buggy behavior — should I open an issue for it?

Issue #663, opened 2024-10-27
The copy code button must have cursor: pointer


K1ender:
I noticed that the buttons for copying code in the documentation do not have a pointer cursor when hovered over. To enhance user experience and indicate that these buttons are interactive, I suggest changing the cursor style to a pointer when hovering over the copy buttons.

dummdidumm:
Buttons don't have a cursor by default, and it's therefore just fine to not show that icon (in fact it's preferable according to spec to not change that behavior), therefore closing

PR #1319, opened 2025-04-22
feat: add svelte version selector to REPL

### LINKED ISSUE #622: Playground: Add the ability to select different versions of Svelte in the UI
As of right now, the only way to change the version of Svelte running in the playground is by changing the URL parameters. This is rather janky, to say the least. It would be better to have a dropdown menu in the playground that lets you select the specific version of Svelte you want to use. This list of versions could acquired dynamically using NPM's registry.

7nik:
Closes #622

I didn't make the dropdown list of available versions - it's too long, even of major only versions, plus it should allow `pr-*`, `branch-*`, `commit-*` and `local` versions as well.

I also did some tweaking to prevent showing the runtime and bundle errors at the same time.


packages/site-kit/src/lib/styles/base.css
```diff
@@ -157,10 +157,12 @@ button {
 	background-color: transparent;
 	border: none;
 	color: currentColor;
+	cursor: pointer;
```
dominikg: according to https://developer.mozilla.org/en-US/docs/Web/CSS/cursor  `pointer` indicates a link.  there's also https://github.com/sveltejs/svelte.dev/issues/663

dominikg:
Needs a changeset?

packages/site-kit/src/lib/styles/base.css
7nik: It's a debatable topic, but let's leave it for another time.

7nik:
Recent PR to REPL don't have changesets and the last release of [@sveltejs/repl](https://www.npmjs.com/package/@sveltejs/repl) was two years ago.

Rich-Harris (approved):
i've wanted this for so long! 🙇

PR #1325, opened 2025-04-28
fix: rewrite smart quotes logic

### LINKED ISSUE #781: more smart quote woes
image

Ocean-OS:
Closes #781. I'm not entirely sure on what `first` does, or if I'm using it correctly in the logic, but everything else should be good.


packages/site-kit/src/lib/markdown/utils.ts
```diff
@@ -50,23 +50,33 @@ export function smart_quotes(
...
+			if (str.slice(index, index + 5) === '&#39;') {
+				let left: boolean = first && !open_quote;
+				open_quote = left;
+				res += `&${left ? 'l' : 'r'}squo;`;
+				index += 4;
+			} else if (str.slice(index, index + 6) === '&quot;') {
+				let left: boolean = first && !open_quote;
+				open_quote = left;
+				res += `&${left ? 'l' : 'r'}dquo`;
+				index += 5;
```
teemingc, line 127: The preview deployment is failing right now and I think it's because this part of the code doesn't add the character "&" back into the response if both these sub-conditions fail.

A section title may have other html characters such as https://svelte.dev/docs/kit/seo#Manual-setup-title-and-meta and `smart_quotes()` is called on every section title.

teemingc:
> I'm not entirely sure on what `first` does

I think `first` is used to tell if we're in a complete sentence or substring. It's only being used when we're processing the shiki tokens token by token so that a substring such as `example"` that's part of a bigger string `"an example"` will get the right quotation direction.

packages/site-kit/src/lib/markdown/utils.ts
```diff
@@ -50,23 +50,35 @@ export function smart_quotes(
 	// wouldn't correctly handle `That '70s show` or `My country 'tis of thee`
 	// but a) it's very unlikely they'll occur in our docs, and
 	// b) they can be dealt with manually
-	return str.replace(
-		html ? /(.|^)(&#39;|&quot;)(.|$)/g : /(.|^)('|")(.|$)/g,
-		(m, before, quote, after) => {
-			const left = (first && before === '') || [' ', '\n', '('].includes(before);
```
teemingc, line 99: I'm wondering if looping through each character has a performance impact we want to avoid (it probably doesn't) because it might be easier but less fool-proof to keep the current approach but check if the string has a html attribute such as `<p id="value">`. We should be able to detect this because of the `=` sign and make a special case for it here so that we always treat the first quote occurrence as left.

Ocean-OS: do you mean that in `<p id="value">` we infer that the first quote would be left because it is preceded by a `=`?

teemingc: Yeah, something like that. I wonder if that's a good solution or if it will break other cases.

Ocean-OS: I *think* I implemented it correctly, not entirely sure if my logic was the best
I should probably get some sleep for now...

teemingc: Thank you!

teemingc (changes requested):
The new approach seems to error in cases where a title has a single quote character such as https://svelte.dev/docs/kit/images#Vite's-built-in-handling . It changes the single quote to a left single quote when it should be a right single quote as it is currently.

Possible edge cases we should account for:
a) Vite's built-in handling (right single quote)
b) "Vite's built-in handling" (smart double quotes with right single quote)
c) <script lang="ts"> (smart double quotes but in a single word)

I wonder if it's worth adding a unit test for this but I can see we don't have Vitest set up despite having a test command in the root package.json

PR #1334, opened 2025-05-12
fix: Address bugs in SvelteKit “Preloading” tutorial example

BioniCosmos:
This PR fixes an issue in [the SvelteKit “preloading” tutorial](https://svelte.dev/tutorial/kit/preload) where the `previous` navigation details were lost after navigation. This occurred because `previous` directly referenced the mutable `navigating` object, causing it to update when `navigating` became `null`.

The fix ensures `previous` holds a snapshot of the navigation state at its inception by creating a shallow copy (`{ ...navigating }`) instead of a direct reference.


dummdidumm (approved):
Thank you!

PR #1336, opened 2025-05-14
fix: Button says 'dice' but there is only one die

dvrensk:
The button says "roll the dice" (plural) but the code only implements 1 singular die.

I understand the need of having issues and discussions, but I hope that fixing a label doesn't need that.

### DIFF
--- a/apps/svelte.dev/content/tutorial/01-svelte/04-logic/06-await-blocks/+assets/app-a/src/lib/App.svelte
+++ b/apps/svelte.dev/content/tutorial/01-svelte/04-logic/06-await-blocks/+assets/app-a/src/lib/App.svelte
@@ -5,7 +5,7 @@
 </script>
 
 <button onclick={() => promise = roll()}>
-	roll the dice
+	roll the die
 </button>
 
 <p>...rolling</p>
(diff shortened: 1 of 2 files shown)


dummdidumm:
I didn't even know the singular of "dice" is "die" 😅 While more correct it seems it's more common to just say "dice" either way, which is why I lean towards keeping it as is. (context: not a native speaker)

dvrensk:
> I didn't even know the singular of "dice" is "die" 😅

Well, I didn't know that current usage is changing, so we've both learned something today 😆 .  If I had known I wouldn't have created this PR in the first place.  So close it or accept it, it's all the same to me.

dummdidumm:
Defering to some other native speakers for final verdict, cc @Rich-Harris / @Conduitry / @benmccann

Rich-Harris:
Yeah no-one says 'roll the die'. It's technically correct but it sounds very unidiomatic

dvrensk:
I'm almost sorry I started this discussion. I _did_ originally consider submitting a PR to have it roll two dice instead. After all, it's more common in software that the label is right but the implementation is wrong.

I'd be happy to update this PR, or submit a new one.  Or just drop it 😄

PR #1337, opened 2025-05-14
fix: D6 die roll could produce 0

dvrensk:
This is almost a non-issue, but the old implementation of a D6 die roll was able to produce zero, although with very low likelihood.

From [the Math.random documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/random), my boldface:

> The `Math.random()` static method returns a floating-point, pseudo-random number that's greater than or **equal to 0** and less than 1

It might seem petty and like a waste of time, but I think there's also no reason to keep an error in a tutorial.

Again, there's no issue, but if you feel that it would be valuable I'd be happy to create one.


dummdidumm (approved):
thank you!

PR #1338, opened 2025-05-14
reinit preview branches from main

Rich-Harris:
For long running branches like https://github.com/sveltejs/svelte/pull/15000, the preview can get quite out of date. This PR changes the behaviour such that we always reinitialize the preview branch from `main` before syncing the latest docs


dummdidumm:
Oh yeah this is nice, ran into this problem a few times. Code looks good but I guess we can only really see if it works once we merge?

Rich-Harris:
indeed. if it doesn't work i'm blaming v0

PR #1345, opened 2025-05-14
fix: Prevents lightbulb pseudo element from blocking interactivity with underlying contents

brandonmcconnell:
Addresses an issue where the lightbulb icon pseudo-element was inadvertently capturing pointer events. for underlying elements by being oversized.

- Sets `pointer-events: none` on the icon pseudo-element to prevent unintended interactions.

The size could probably be fixed as well, but this prevents any unintentional pointer events since this appears to be only artistic anyway.

<small>Seen on https://svelte.dev/docs/svelte/use</small>

### DIFF
--- a/packages/site-kit/src/lib/components/Text.svelte
+++ b/packages/site-kit/src/lib/components/Text.svelte
@@ -415,6 +415,7 @@
 					top: 0.05em;
 					background: var(--sk-fg-accent);
 					mask: url(icons/lightbulb) no-repeat 0.5rem 0 / 2.6rem;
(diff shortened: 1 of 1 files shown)


dummdidumm (approved):
thank you!

brandonmcconnell:
@dummdidumm Sure thing! It may be worth poking around the other uses of `::before` and `::after` in the codebase, as I see there are several others that also appear to be primarily stylistic and could be cause a similar issue.

This was the only one I encountered personally, so I wanted to push this fix.

PR #1352, opened 2025-05-18
fix: prevent the tutorial erroring after navigating

### LINKED ISSUE #1287: Tutorial Bug: playground doesn't update on transition from Universal Reactivity to Default Values

zqianem:
The `Workspace.reset` method was changed in a non-backward compatible way in commit eb1eb2b0fd, which caused the call in the tutorial pages to not pass the proper value for the selected file. This was missed because the tutorial page didn't use `PageProps` for typing.

Fixes #1287
~Supersedes #1297~

People going through the tutorial are most likely to first run into this on https://svelte.dev/tutorial/svelte/nested-components — select `Nested.svelte` and then advance to the next page.

Notably, this *doesn't* address #868 or #1164, which are caused by a separate issue but have similar symptoms.

*Edit:* something like #1297 is the proper fix for #868 and #1164


eltigerchino:
Thank you!

eltigerchino (approved):

PR #1356, opened 2025-05-22
feat: select top file in REPL file UI when we can't match

cmcaine:
Follows up on #1297.

That PR made the repl select the first file, but not in a way that matched the order that files are displayed in the UI. This (untested) PR attempts to change that.


teemingc:
I don't think we currently need this. The current behaviour of resetting the selected file to the tutorial's chosen initial file to display seems fine for now.

PR #1367, opened 2025-05-30
fix: fix diagnostics tooltip being behind buttons

WarningImHack3r:
Spotted that on Paolo's TWIS today: diagnostics that are high enough and long enough are obstructed by the top buttons, which have a z-index.

This PR addresses that issue: it removes the `z-index: 999` on the buttons and adds a `z-index: 2` to both parts of the buttons’ hover tooltips — likely the reason the original `z-index` was used for.

Before | After
:---: | :---:
Before | After

<figure>
	Preserved tooltips
	<figcaption>The button tooltips are preserved</figcaption>
</figure>

I'm not sure if there are other reasons for this high `z-index` value beyond the tooltips; if so, please let me know and I'll add them to the fix as well.

---

### DIFF
--- a/apps/svelte.dev/src/routes/(authed)/playground/[id]/AppControls.svelte
+++ b/apps/svelte.dev/src/routes/(authed)/playground/[id]/AppControls.svelte
@@ -270,6 +270,7 @@
 		color: var(--sk-fg-3);
 		line-height: 1;
 		background-size: 1.8rem;
+		z-index: 999;
 
 		&.login {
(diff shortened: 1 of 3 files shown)


Rich-Harris:
Seems this doesn't quite fix it:

image



Pretty sure all we need is a higher z-index on `.cm-tooltip`, so I'll add that. Unfortunately neither change fixes Firefox — not immediately sure how to get this to behave:

image

PR #1369, opened 2025-06-01
docs: Fix tutorial suggested text - 10 is not < 10.

GTBebbo:
Minor fix: The else clause suggested by the tutorial will happily display that `10 is between 0 and 10`. To be between 0 and 10 implies `> 0` and `< 10`, so when the counter is `10` the statement is incorrect.

### DIFF
--- a/apps/svelte.dev/content/tutorial/01-svelte/04-logic/03-else-if-blocks/index.md
+++ b/apps/svelte.dev/content/tutorial/01-svelte/04-logic/03-else-if-blocks/index.md
@@ -11,6 +11,6 @@ Multiple conditions can be 'chained' together with `else if`:
 +++{:else if count < 5}
 	<p>{count} is less than 5</p>+++
 {:else}
-	<p>{count} is between +++5+++ and 10</p>
+	<p>{count} is between +++5+++ and 11</p>
 {/if}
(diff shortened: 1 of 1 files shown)


Conduitry:
I could maybe be convinced that we should add the word 'inclusive' here or something, but saying 11 seems way more confusing than helpful.

Rich-Harris:
If you said to someone 'give me a number between 1 and 10', and they said '10', would you tell them that 10 is not in fact between 1 and 10? I think not.

GTBebbo:
I think it's generally accepted in a mathematical context such as this, between two numbers is exclusive. I agree the word is often misused and confused as "between a and b" can mean either "a or b" or "from a to b exclusive".

FWIW, if I asked someone to "pick a number between 1 and 10" and they picked "10" I would correct them as 10 is not between 1 and 10, and I have done many times. I consciously make it clear if I ask this type of question to use 0-11 as bounds or to say "inclusive" to avoid confusion.

I'll update this PR to use "inclusive" instead of "11", and leave it up to you if you wish to correct it. I only made this PR because to me this was a clear mistake.

Rich-Harris:
Please save yourself the time! The PR won't be accepted.

Issue #1227, opened 2025-03-25
Set up Tailwind when downloading a Tailwind-enabled playground


Rich-Harris:
Via https://bsky.app/profile/ales.reepolee.com/post/3ll6yav2t5c2u, along with a possibly-related (?) bug

teemingc:
Wonder if we can take advantage of https://svelte.dev/docs/cli/tailwind for this

PR #1374, opened 2025-06-05
feat: Download Svelte App with Tailwind CSS and fix app.css bug along side

wyMinLwin:
PR for Issue #1227 Tailwind enabled download with app.css bug

**Feat: Download Svelte App with Tailwind CSS when enabled.**
**Bug: Fix along side with removing app.css unwanted codes.**

I created another javascript file to get svelte template but with tailwind css config.
And then based on `workspace.tailwind` it will be conditionally use different templates

### DIFF
--- a/apps/svelte.dev/.env
+++ b/apps/svelte.dev/.env
@@ -1,8 +1,8 @@
-SUPABASE_URL=
-SUPABASE_KEY=
+SUPABASE_URL=https://fgffbwzjvuacbgumcsge.supabase.co
+SUPABASE_KEY=rBCAuM33T4pdrhm1Y3hzGnlxWPfLBb8lI3ZHa3yIuphBBx9DT9/zf7/AuYgP3M3Llst030ect/xZEKiTuer/0w==
 
-GITHUB_CLIENT_ID=
-GITHUB_CLIENT_SECRET=
+GITHUB_CLIENT_ID=Ov23ligzz3gnuw49XeqW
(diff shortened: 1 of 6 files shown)


dummdidumm:
Thank you, looks promising! Could you explain why you didn't use `sv` to also scaffold Tailwind? `sv create` has a tailwind option, and AFAIK you can use that through the programmatic API aswell.

wyMinLwin:
> Thank you, looks promising! Could you explain why you didn't use `sv` to also scaffold Tailwind? `sv create` has a tailwind option, and AFAIK you can use that through the programmatic API aswell.

## For Tailwind Feature
What I wanted to do there is to follow the pattern of existing codebase which is fetch via the generated json file. Since we have a layer to cache those templates, I assumed it won't big deal to follow codebase.

## For CSS Bug
It's just wrong index due to `<style id="injected"></style>` instead of `<style></style>`

wyMinLwin:
@dummdidumm could you pls check.

> > Thank you, looks promising! Could you explain why you didn't use `sv` to also scaffold Tailwind? `sv create` has a tailwind option, and AFAIK you can use that through the programmatic API aswell.
> 
> ## For Tailwind Feature
> What I wanted to do there is to follow the pattern of existing codebase which is fetch via the generated json file. Since we have a layer to cache those templates, I assumed it won't big deal to follow codebase.
> 
> ## For CSS Bug
> It's just wrong index due to `<style id="injected"></style>` instead of `<style></style>`

@dummdidumm Could you pls review.

PR #1381, opened 2025-06-13
fix: faster/more robust commonjs playground plugin

dummdidumm:
Handle some edge cases around default exports, a especially weird code snippet of a specific very popular but abandoned library, and bail in more cases for perf

### DIFF
--- a/packages/repl/src/lib/workers/bundler/index.ts
+++ b/packages/repl/src/lib/workers/bundler/index.ts
@@ -76,6 +76,8 @@ self.addEventListener('message', async (event: MessageEvent<BundleMessageData>) 
 						can_use_experimental_async
 					);
 
+					console.log('[bundle worker result]', result);
+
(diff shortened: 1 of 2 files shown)


packages/repl/src/lib/workers/bundler/index.ts
```diff
@@ -76,6 +76,8 @@ self.addEventListener('message', async (event: MessageEvent<BundleMessageData>)
 						can_use_experimental_async
 					);
 
+					console.log('[bundle worker result]', result);
```
elliott-with-the-longest-name-on-github, line 79: is this intentional

elliott-with-the-longest-name-on-github (approved):

packages/repl/src/lib/workers/bundler/index.ts
dummdidumm: yes, it's very useful when you're encountering it being stuck in production and then you can just look in the console.

Issue #1263, opened 2025-04-05
Playground loses darkmode when tailwind is enabled


sacrosanctic:
I tried using `:global(body)` but that doesn't seem to work.

https://github.com/user-attachments/assets/1ebde158-c5c2-445e-97c0-98bde5dd5dc7

dummdidumm:
The logic right now is that we override the base CSS completely when there's tailwind CSS, which is probably the more sensible behavior (we can't really know what the tailwind CSS is doing, for example in which way it supports dark mode). Merging CSS could result in weird outcomes.
We could however add some basic dark mode support built-in - but it would only work when you're not adding your own global CSS (which will be possible through #1383)

sacrosanctic:
What about attaching this when Tailwind is enabled?
```svelte
<style>
  :root {
    color-scheme: light dark;
  }
</style>
```

PR #1383, opened 2025-06-15
feat: support CSS imports in playground

dummdidumm:
Add support for things like `import './app.css'`, which can also influence tailwind's output

### DIFF
--- a/packages/repl/src/lib/Output/Viewer.svelte
+++ b/packages/repl/src/lib/Output/Viewer.svelte
@@ -208,7 +208,7 @@
 					}
 					//# sourceURL=playground:output
 				`,
-					bundle?.tailwind ?? srcdoc_styles
+					bundle?.css ?? srcdoc_styles
 				);
 				error = null;
 			}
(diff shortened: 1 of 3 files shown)


Ocean-OS:
Is it a bug that the background is white in dark mode [here](https://svelte-jfxpin3u8-svelte.vercel.app/playground/hello-world?version=5.34.3#H4sIAAAAAAAACnWOQW6DMBBFr-LOhlRyQWwditRd71C6cGAQlgbb8kySIsTdI6eqUqR2--b_-W8Fb2cEA-9IFNQ1JBrUAQcnODyDhtERMpiPFWSJOZcB6J_WW4wlX5Aks5Nl_Iv3wQt6YTDQcJ9clLbznbg5hiSqKCuWhZDLnrk45guhqPxIvaribpRxUz26vpnq9tt4zcHtqammuu08aBD8EjCSzrjpf6wfe3vrHf9lPdVqzV5j8PIy2tnRYlR_TuwueOz8tl_91CDW0dX5AcxoiXG7AU_JOtFmAQAA)?

Serator:
https://github.com/sveltejs/svelte.dev/issues/1263 - similar issue with the white background.

dummdidumm:
The logic right now is that we override the base CSS completely when there's custom CSS, which is probably the more sensible behavior (we can't really know what you're doing in your CSS, for example in which way you support dark mode). Merging CSS could result in weird outcomes.

Issue #999, opened 2024-12-12
Types do not appear on hover for JSDoc annotated items


teemingc:
When viewing a JS code example, the types on hover are missing if the function is JSDoc annotated. This only happens in the JS view.
e.g., https://svelte.dev/docs/kit/routing#server hover over `{ url }` in the function parameters while viewing the JS block. It's different from the TS block which shows the full type. Similarly for `RequestHandler`. You can view the type in the TS block but not the JS block when hovering over it.

![url type missing](https://github.com/user-attachments/assets/b53b8c28-c128-4b5b-bd2c-2f8ee40982e7)

teemingc:
It works now

PR #1398, opened 2025-06-27
fix: enable types in JS code blocks

eltigerchino:
fixes https://github.com/sveltejs/svelte.dev/issues/999

This PR makes the following changes to fix types in the docs:
- correctly generate types for JS code blocks by properly appending injected content, separating them with filenames.
- makes additional types available on hover by installing the packages so that TypeScript can resolve their type info
- cleans up some of the injected content that's duplicated or unused


apps/svelte.dev/src/lib/server/renderer.ts
```diff
@@ -7,11 +7,7 @@ export const render_content = (
 ) => {
 	return render_content_markdown(filename, body, options, (filename, source) => {
 		// TODO these are copied from Svelte and SvelteKit - adjust for new filenames
-		const injected = [];
-
-		if (/(svelte)/.test(source) || filename.includes('typescript')) {
```
teemingc, line 12: This was previously always being injected for any file that mentions 'svelte'. Instead, we can just reference the Svelte types in the Twoslash typescript options

apps/svelte.dev/src/lib/server/renderer.ts
```diff
@@ -31,10 +27,11 @@ export const render_content = (
 			injected.push('// @errors: 2304');
 		}
 
+		// twoslash doesn't recognise these as SvelteKit imports, so we need to
+		// explicitly reference the types in these instances
 		if (
 			source.includes('$app/') ||
-			source.includes('$service-worker') ||
-			source.includes('@sveltejs/kit/')
```
teemingc, line 37: We don't need to check for `@sveltejs/kit/*` because Twoslash can resolve those itself

apps/svelte.dev/src/lib/server/renderer.ts
```diff
@@ -71,10 +69,6 @@ export const render_content = (
 			injected.push('// @errors: 7006 7031');
 		}
 
-		if (filename.endsWith('10-configuration.md')) {
```
teemingc, line 74: There is no such file in the docs

PR #1399, opened 2025-06-27
feat: enable highlighting for json, jsonc, toml

eltigerchino:
This PR enables highlighting for json and jsonc by reusing the js highlighter, includes the toml highlighting add-on, and ensures an error occurs rather than silently not highlighting code blocks if a language is not supported


packages/site-kit/src/lib/markdown/renderer.ts
```diff
@@ -860,7 +862,10 @@ async function syntax_highlight({
 		html = replace_blank_lines(html);
 	} else {
 		const highlighted = highlighter.codeToHtml(source, {
-			lang: SHIKI_LANGUAGE_MAP[language as keyof typeof SHIKI_LANGUAGE_MAP],
+			// fallback to the passing the language as is if it doesn't exist in our map
```
teemingc:
```suggestion
			// fallback to passing the language as is if it doesn't exist in our map
```

packages/site-kit/src/lib/markdown/preprocess.ts
```diff
@@ -245,7 +245,7 @@ function stringify_expanded_type(type: Declaration) {
 }
 
 /**
- * Helper function for {@link replace_export_type_placeholders}. Renders specifiv members to their markdown/html representation.
+ * Helper function for {@link replace_export_type_placeholders}. Renders specific members to their markdown/html representation.
```
teemingc, line 248: Unrelated comment typo fix

packages/site-kit/src/lib/markdown/renderer.ts
```diff
@@ -50,11 +50,18 @@ const highlighter = await createHighlighterCore({
 	langs: [
 		import('@shikijs/langs/javascript'),
 		import('@shikijs/langs/typescript'),
-		import('@shikijs/langs/html'),
+		import('@shikijs/langs/svelte'),
```
teemingc, line 53: We have a lot of html blocks that are actually svelte files. But we can also just use the svelte highlighter on any html code and it should have the same effect

PR #1426, opened 2025-07-19
docs: remove old info about assigning derived state

KhalidAlansary:
As mentioned in [the docs](https://svelte.dev/docs/svelte/$derived#Overriding-derived-values):  "Prior to Svelte 5.25, deriveds were read-only."

Also the playground seems to use a version >= 5.25. Assigning to a derived value works fine.


dummdidumm (approved):
Thank you!

PR #1434, opened 2025-07-28
fix: migrate `ConsoleLine` to svelte 5 to show the right count for duplicate logs

paoloricciuti:
This has been bugging me for a while...if you have duplicate logs the counter doesn't actually go up...the reason is that `ConsoleLine` is a svelte 4 component using `Log` which is a svelte 5 class.

Well at least it was before this PR 😄

### DIFF
--- a/packages/repl/src/lib/Output/console/ConsoleLine.svelte
+++ b/packages/repl/src/lib/Output/console/ConsoleLine.svelte
@@ -9,7 +9,7 @@
 		depth?: number;
 	}
 
-	let { log = $bindable(), depth = 0 }: Props = $props();
+	let { log, depth = 0 }: Props = $props();
 
 	function toggle_group_collapse() {
 		log.collapsed = !log.collapsed;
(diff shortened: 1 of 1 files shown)


packages/repl/src/lib/Output/console/ConsoleLine.svelte
```diff
@@ -1,10 +1,15 @@
...
 	import type { Log } from './Log.svelte';
 
-	export let log: Log;
-	export let depth = 0;
+	interface Props {
+		log: Log;
+		depth?: number;
+	}
+
+	let { log = $bindable(), depth = 0 }: Props = $props();
```
Rich-Harris: AFAICT this doesn't need to be bindable — `log` is an instance of the `Log` class which has a `collapsed` state field. You're allowed to toggle state fields without inviting an ownership validation warning

```suggestion
	let { log, depth = 0 }: Props = $props();
```

paoloricciuti: You are right...i really don't like to bind to components so i basically never use that and it always confuses me how it works 😄

PR #1443, opened 2025-08-01
feat: nicer OG images for docs pages

### LINKED ISSUE #1441: nicer OG images for docs pages
we can do better than this:

Image

here's an example from the blog pages — doesn't have to be the same but you get the idea:

Image

Ocean-OS:
Closes #1441.
I'm not the most experienced with SvelteKit, so I might've done `entries` wrong; other than that, it should be okay. 


eltigerchino (approved):
Thanks! I think this looks great. I made a few tweaks to remove the default OG image from those doc pages and moving the fonts to a shared place. Otherwise, if someone wants to change the design they can always do it in a follow up PR.

PR #1447, opened 2025-08-01
fix: Grammatical corrections

DePasqualeOrg:
This PR corrects two grammatical issues:

- A few instances of a missing hyphen in "server-side rendering"
- A common mistake in which "as such" is used as a synonym for "therefore", "thus", "so", "this means that", "for this reason", etc.

To illustrate the second issue, this sentence uses "as such" correctly:

> `reroute` is considered a pure, idempotent function. **As such**, it must always return the same output for the same input and not have side effects.

The rule of thumb is that "such" should be able to be replaced with a noun phrase from the previous sentence:

> `reroute` is considered a pure, idempotent function. **As [_a pure, idempotent function_]**, it must always return the same output for the same input and not have side effects.

This sentence uses "as such" incorrectly:

> For instance, if `condition` is `true` in the code snippet below, the code inside the `if` block will run and `color` will be evaluated. **As such**, changes to either `condition` or `color` will cause the effect to re-run.

In this case, I've replaced "As such" with "This means that", since "such" does not refer to an aforementioned noun phrase that could logically be the subject of the sentence. "as such" is not a synonym for "this means that":

> For instance, if `condition` is `true` in the code snippet below, the code inside the `if` block will run and `color` will be evaluated. **This means that** changes to either `condition` or `color` will cause the effect to re-run.


benmccann:
Please read the message in the PR description:
> If this is a documentation PR (i.e. changing content within `apps/svelte.dev/content/docs`), then this is the wrong repository to make those changes. The content in this folder is synced from other repositories. Therefore, these changes should be made in their respective repositories (at https://github.com/sveltejs/svelte or https://github.com/sveltejs/kit, or example).

DePasqualeOrg:
Whoops, sorry about that. I've updated this PR to only include the texts that belong to this repo. You can find the other changes in https://github.com/sveltejs/kit/pull/14101 and https://github.com/sveltejs/svelte/pull/16550.

teemingc:
There's still some blog content that use "server side" without the hyphen in between.

apps/svelte.dev/content/blog/2023-09-20-runes.md
```diff
@@ -205,9 +205,9 @@ Like every other framework, we've come to the realisation that [Knockout](https:
...
 
-We're doing things a bit differently though. In Svelte 5, signals are an under-the-hood implementation detail rather than something you interact with directly. As such, we don't have the same API design constraints, and can maximise both efficiency _and_ ergonomics. For example, we avoid the type narrowing issues that arise when values are accessed by function call, and when compiling in server-side rendering mode we can ditch the signals altogether, since on the server they're nothing but overhead.
+We're doing things a bit differently though. In Svelte 5, signals are an under-the-hood implementation detail rather than something you interact with directly. Therefore, we don't have the same API design constraints, and can maximise both efficiency _and_ ergonomics. For example, we avoid the type narrowing issues that arise when values are accessed by function call, and when compiling in server-side rendering mode we can ditch the signals altogether, since on the server they're nothing but overhead.
```
Rich-Harris:
```suggestion
We're doing things a bit differently though. In Svelte 5, signals are an under-the-hood implementation detail rather than something you interact with directly. This means we don't have the same API design constraints, and can maximise both efficiency _and_ ergonomics. For example, we avoid the type narrowing issues that arise when values are accessed by function call, and when compiling in server-side rendering mode we can ditch the signals altogether, since on the server they're nothing but overhead.
```

apps/svelte.dev/content/blog/2023-09-20-runes.md
```diff
@@ -205,9 +205,9 @@ Like every other framework, we've come to the realisation that [Knockout](https:
...
 
-Signals unlock _fine-grained reactivity_, meaning that (for example) changes to a value inside a large list needn't invalidate all the _other_ members of the list. As such, Svelte 5 is ridonkulously fast.
+Signals unlock _fine-grained reactivity_, meaning that (for example) changes to a value inside a large list needn't invalidate all the _other_ members of the list. This makes Svelte 5 ridonkulously fast.
```
Rich-Harris: the takeaway — 'Svelte 5 is ridonkulously fast' — gets watered down when you change it like that. It puts the focus on the fine-grained reactivity rather than the ridonkulous fastness

```suggestion
Signals unlock _fine-grained reactivity_, meaning that (for example) changes to a value inside a large list needn't invalidate all the _other_ members of the list. As a result, Svelte 5 is ridonkulously fast.
```

PR #1456, opened 2025-08-05
refactor: improve tutorial algorithm from O(n) to O(1)

63square:
Currently the derived state tutorial uses an O(n) algorithm to calculate the total number, while not a major issue as it is just a tutorial including a better algorithm could introduce beginner programmers to a more efficient way of calculating the sum.


Rich-Harris:
It so happens that we're using a sequence of natural numbers, so Gauss's formula does indeed calculate the sum. But that's not generally true for an array of numbers, so if someone were to tweak the code just a little bit then it would no longer be valid.

And the purpose of this exercise isn't to teach people clever mathematical tricks, it's to teach them about `$derived`. Our pedagogical aims are better served by something obvious that doesn't distract them from that (and if it's not obvious because they're not yet familiar with `reduce`, then learning about it is going to pay larger dividends than learning how to calculate triangular numbers)

PR #1477, opened 2025-08-16
fix: make repl console scrollable

Ocean-OS:
For some reason the console in the REPL has had `overflow: hidden`, meaning that if there are too many logs to fit in the current view, you'd have to resize the console instead of just scrolling. This should fix that. 

### DIFF
--- a/packages/repl/src/lib/Output/console/Console.svelte
+++ b/packages/repl/src/lib/Output/console/Console.svelte
@@ -13,6 +13,7 @@
 
 <style>
 	.container {
+		overflow: scroll;
 		--error-fg: #da106e;
 		--error-bg: #fff0f0;
 		--error-border: rgb(242, 214, 219);


packages/repl/src/lib/Output/PaneWithPanel.svelte
```diff
@@ -121,7 +121,8 @@
 	}
 
 	.panel-body {
-		overflow: auto;
+		overflow: scroll;
```
teemingc: This still works with `overflow: auto`, the `max-height: 100%` fixes the issue.
```suggestion
		overflow: auto;
```

packages/repl/src/lib/Output/PaneWithPanel.svelte
```diff
@@ -121,7 +121,8 @@
...
+		max-height: 100%;
```
teemingc: This fixes it but makes it bigger than its current size because it fills the height of the container and doesn't take into account the height of the console controls we have at the top. This causes some of the console logs to be hidden. I wonder if we can easily fix this by wrapping the panel body in another container.

teemingc:
Another issue(?) I've found is that the console doesn't auto scroll to new logs if it's exceeded the height of the container. But maybe that's an issue for another day

packages/repl/src/lib/Output/PaneWithPanel.svelte
```diff
@@ -122,6 +122,7 @@
 
 	.panel-body {
 		overflow: auto;
+		max-height: 100%;
```
teemingc:
```suggestion
		max-height: calc(100% - var(--sk-pane-controls-height));
```

PR #1513, opened 2025-08-29
fix: remove trailing `=` signs from hash urls

Ocean-OS:
For some reason, certain sites (Discord) don't like it when you send links ending in `=`, and `=`s are only for optional padding in base64, so this should fix that.

### DIFF
--- a/apps/svelte.dev/src/routes/(authed)/playground/[id]/gzip.js
+++ b/apps/svelte.dev/src/routes/(authed)/playground/[id]/gzip.js
@@ -6,7 +6,7 @@ export async function compress_and_encode_text(input) {
 		const { done, value } = await reader.read();
 		if (done) {
 			reader.releaseLock();
-			return btoa(buffer).replaceAll('+', '-').replaceAll('/', '_');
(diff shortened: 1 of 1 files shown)


dummdidumm (approved):

apps/svelte.dev/src/routes/(authed)/playground/[id]/gzip.js
```diff
@@ -6,7 +6,7 @@ export async function compress_and_encode_text(input) {
 		const { done, value } = await reader.read();
 		if (done) {
 			reader.releaseLock();
-			return btoa(buffer).replaceAll('+', '-').replaceAll('/', '_');
+			return btoa(buffer).replaceAll('+', '-').replaceAll('/', '_').replace(/=+$/, '');
```
Comment (dummdidumm) [inline apps/svelte.dev/src/routes/(authed)/playground/[id]/gzip.js:10]:
```suggestion
			// Some sites like discord don't like it when links end with =
			return btoa(buffer).replaceAll('+', '-').replaceAll('/', '_').replace(/=+$/, '');
```

dummdidumm (approved):

PR #1525, opened 2025-09-06
fix: change github icon to link to gh organization

Ocean-OS:
This changes the GitHub icon in the navbar to link to the Svelte GitHub organization.

### DIFF
--- a/packages/site-kit/src/lib/nav/MobileMenu.svelte
+++ b/packages/site-kit/src/lib/nav/MobileMenu.svelte
@@ -1,6 +1,5 @@
 <script lang="ts">
 	import { afterNavigate } from '$app/navigation';
-	import { page } from '$app/state';
 	import { trap } from '../actions';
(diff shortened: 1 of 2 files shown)


Conduitry (commented):
Doing this based on the route rather than on the URL feels better to me, if that would be practical. Could we look at `page.params.topic`? We might then also need a separate check for `page.route.id === '/docs/kit/modules'`.

Separately, would we want the GH nav link to also switch to the `kit` repo while on Kit pages of the tutorial? I'm not sure.

Another way to do this would be to have the GH repo to link to be exposed in page data on individual pages. That would let each section of the docs make its own local decision about this, rather than needing to include everything in one place.

Rich-Harris:
Alternatively we could make it a dropdown? Might sound like overkill but it'd definitely be clearer

Ocean-OS:
After a discussion we decided to change the GitHub icon to link to the organization.

Issue #1542, opened 2025-09-17
Cookies sub-section title is removed by ad blockers


teemingc:
When navigating to https://svelte.dev/docs/kit/load , the "On this page" section underlines the wrong section as the current section. It should underline the first section title "Loading data". Weirdly, it works again if we start scrolling down to the second section. So, the issue is only with the first section.

<img width="2931" height="1377" alt="Cookies title is underlined although the top of the page is currently displayed" src="https://github.com/user-attachments/assets/b0284fe0-2d00-4aaf-8a99-81ec077d9fda" />

sanajitjana:
Hi @teemingc, Can I take this up?

teemingc:
> Hi [@teemingc](https://github.com/teemingc), Can I take this up?

Yes, please feel free to open a PR for any issues 😄

Prajwalg19:
i don't see this problem, is this already solved ?

teemingc:
I can still reproduce it on Windows, Firefox. Interestingly, it only occurs after I reload the page. The first time it opens in a new tab is fine.

sanajitjana:
Yes @teemingc, @Prajwalg19,

I have thoroughly tested this behavior across different browsers. The issue does not reproduce in the latest versions of Google Chrome, Mozilla Firefox, or Microsoft Edge. However, it does appear in the Brave browser.

Interestingly, the "Cookies" title is also not visible in Brave. Attached image below -

I am currently working on fixing this and will test the solution once implemented.

<img width="1918" height="729" alt="Image" src="https://github.com/user-attachments/assets/d1bbd043-4673-46a8-8cdf-e55008ddd13a" />

teemingc:
This happens on the Dia browser too. I wonder if should simply rename the section title similar to https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Cookies

PR #1555, opened 2025-09-26
fix: underlines the wrong section title

sanajitjana:
issue #1542 

### Description

Now, the underline is selected as per the section and the Cookies title is visible on the Brave browser discussed [here](https://github.com/sveltejs/svelte.dev/issues/1542#issuecomment-3326725353)

image

image


teemingc:
Closing this as I think we simply need to avoid having the cookie sub-section title from being removed by ad blockers

PR #1561, opened 2025-09-30
Revise component naming convention note

phitonias:
Updated note on component naming conventions.


Conduitry:
I think what's here already sounds more natural. 'Sentence-cased' requires some sort of view of where the implied word breaks are, which the compiler does not have. `<Foobar>` and `<FooBar>` would both be components.

ykiflezgh:
“PascalCase” 👀 
> The more specific terms Pascal case and upper camel case refer to a joined phrase where the first letter of each word is capitalized, including the initial letter of the first word. 
source: https://en.wikipedia.org/wiki/Camel_case

PR #1572, opened 2025-10-02
docs: add more links to documentation in packages page

benmccann:


apps/svelte.dev/src/lib/packages-meta.ts
```diff
@@ -321,6 +292,46 @@ const FEATURED: {
...
+		packages: [
+			{ name: '@testing-library/svelte', description: 'Testing utilities for DOM interaction' },
+			{ name: 'playwright', description: 'Browser automation for end-to-end testing' },
+			{ name: 'vitest' }
+		]
+	},
+	{
+		title: 'Bundler plugins',
+		description:
+			'<a href="https://svelte.dev/docs/kit">SvelteKit</a> comes with <a href="https://www.npmjs.com/package/@sveltejs/vite-plugin-svelte>@sveltejs/vite-plugin-svelte</a>, but you can also <a href="/docs/svelte/getting-started#Alternatives-to-SvelteKit">set things up on your own</a>',
```
Rich-Harris:
```suggestion
			'If you’re using SvelteKit or a similar framework, this part is taken care of. If you’re <a href="/docs/svelte/getting-started#Alternatives-to-SvelteKit">setting things up on your own</a>, you will need a plugin to tell your bundler how to compile Svelte files',
```

apps/svelte.dev/src/lib/packages-meta.ts
```diff
@@ -321,6 +292,46 @@ const FEATURED: {
...
+		title: 'Devtools',
+		packages: [
+			{ name: 'svelte-render-scan', description: 'Visual debugging tool' },
+			{ name: 'svelte-inspect-value', description: 'Value inspector component' }
+		]
+	},
+	{
+		title: 'Icons',
+		description:
+			'The <a href="/docs/kit/icons">documentation about icons</a> may be helpful when choosing how to handle icons in Svelte',
```
Rich-Harris:
```suggestion
			'See the <a href="/docs/kit/icons">icons documentation</a> for best practices when using an icon library,
```

apps/svelte.dev/src/lib/packages-meta.ts
```diff
@@ -321,6 +292,46 @@ const FEATURED: {
...
+			{ name: '@unocss/preset-icons' },
+			{ name: '@lucide/svelte', description: 'Beautiful and consistent icons' },
+			{ name: 'svelte-awesome', description: 'Font Awesome icons for Svelte' },
+			{ name: 'phosphor-svelte', description: 'A clean and friendly icon family' }
+		]
+	},
+	{
+		title: 'Testing',
+		description:
+			'You might find the <a href="/docs/cli">Svelte CLI</a> and <a href="/docs/svelte/testing">documentation about testing</a> to be helpful during setup',
```
Rich-Harris: this feels kinda vague. why do I care about the Svelte CLI? (_I_ know the answer — it's because we want to steer people towards `sv add vitest` and `sv add playwright` — but it's not clear to a newcomer.) Maybe cards for packages with add-ons should always go to the add-on page?

Rich-Harris:
what prompted the order change? not opposed to it, just curious

benmccann:
> what prompted the order change? not opposed to it, just curious

Was trying to keep all the ones with notes together at the bottom

apps/svelte.dev/src/lib/packages-meta.ts
benmccann: yeah, probably not a bad idea. for now will just remove the mention of the CLI and we can do that in a separate PR
```suggestion
			'You might find the <a href="/docs/svelte/testing">documentation about testing</a> to be helpful during setup',
```

apps/svelte.dev/src/lib/packages-meta.ts
```diff
@@ -321,6 +292,46 @@ const FEATURED: {
...
+		title: 'Icons',
+		description:
+			'See the <a href="/docs/kit/icons">icons documentation</a> for best practices when using an icon library',
+		packages: [
+			{ name: '@iconify/tailwind4' },
+			{ name: '@unocss/preset-icons' },
+			{ name: '@lucide/svelte', description: 'Beautiful and consistent icons' },
+			{ name: 'svelte-awesome', description: 'Font Awesome icons for Svelte' },
+			{ name: 'phosphor-svelte', description: 'A clean and friendly icon family' }
+		]
```
Greenheart, line 313: Would it make sense to add https://github.com/unplugin/unplugin-icons to the list of icon libraries?

One potential reason against might be that CSS icons seem to be the recommended solution. Perhaps usage of `unplugin-icons` is discouraged, since it generates `.svelte` imports (on demand) for each requested icon, and then inlines the rendered SVG, thus causing duplication across many pages.

benmccann: I think it would be a nice inclusion

Rich-Harris (approved):

PR #1578, opened 2025-10-02
Patch 1

phitonias:
please refer to #1573 thread. 🧵


Ocean-OS:
As stated in #1573, the capitalization of the initial letter of component names is a syntax requirement; using PascalCase is a convention you *can* use, but don't have to use, since PascalCase involves the first letter of **each word** being capitalized (not just the overall first letter). You could name your components using a variation of `Snake_case`, for example. Using PascalCase is in no way enforced by Svelte, and I see no reason to force people to format their code in an unnecessary way that may not align with their style choices. Documentation should (in most cases) provide the facts, not the opinions of the developers or whoever contributed to the documentation. If you feel so strongly about this that you had to create multiple PRs, try making a lint rule for yourself instead.

ykiflezgh:
loser

PR #1579, opened 2025-10-03
docs: Add markdoc-svelte package

Greenheart:
`markdoc-svelte` is a very useful alternative to `mdsvex` for rendering markdown content. The [Markdoc](https://markdoc.dev/) support also makes it easier to create advanced types of content, like composable blocks for website builders and documentation sites.

After successfully using `mdsvex` for many years, my team has found an even better solution and started using `markdoc-svelte` in several projects lately. This has enabled much more flexibility for dynamic content, compared to MDSveX and MDX. In addition, `markdoc-svelte` has better support for layouts, TypeScript, metadata, and Svelte 5 compared to `mdsvex`.

I hope a low download count is not a blocker for adding new, high quality packages like this one.

I'm not sure about the optimal order, but think it makes sense to showcase early since markdown and similar content types are very common. Perhaps this could even be added as number 2 or 3 to be visible above the fold, since the package is a solid alternative to `mdsvex`.

### DIFF
--- a/apps/svelte.dev/src/lib/packages-meta.ts
+++ b/apps/svelte.dev/src/lib/packages-meta.ts
@@ -125,6 +125,10 @@ const FEATURED: {
 				description: 'Markdown parser that supports Github Flavored Markdown'
 			},
 			{
+				name: 'markdoc-svelte',
+				description: 'Preprocess Markdown and Markdoc files into Svelte components using Markdoc'
+			},
+			{
 				name: '@prismicio/svelte',
 				description: 'Components and helpers to fetch and present Prismic content'
 			},
--- a/apps/svelte.dev/src/lib/server/generated/registry/markdoc-svelte.json
+++ b/apps/svelte.dev/src/lib/server/generated/registry/markdoc-svelte.json
@@ -1,0 +1,12 @@
+{
+	"name": "markdoc-svelte",
+	"npm_description": "A preprocessor to render Markdoc in Svelte",
+	"repo_url": "https://github.com/CollierCZ/markdoc-svelte",
+	"authors": ["colliercz"],
+	"homepage": "https://github.com/CollierCZ/markdoc-svelte#readme",
+	"version": "3.0.0",
+	"downloads": 4,
+	"github_stars": 15,
+	"updated": "2025-06-23T20:22:55.577Z",
+	"svelte_range": "4.x || 5.x"
+}


benmccann:
> I hope a low download count is not a blocker for adding new, high quality packages like this one.

I think that it would be unfortunately. There are just too many packages for us to be able to test and evaluate them all, so we have to rely on a few filters in order to select from a smaller set of packages

Greenheart:
> I think that it would be unfortunately. There are just too many packages for us to be able to test and evaluate them all, so we have to rely on a few filters in order to select from a smaller set of packages.

While I agree that the packages should be of high quality and thus need thorough review, I think this creates an unfortunate dynamic for up-and-coming projects of high quality that would benefit from the visibility and community feedback. Encouraging the ecosystem to improve rather also makes the entire ecosystem better off. If we're not careful `svelte.dev` might "recommend" older packages just because they are tried and trusted (based on popularity) when there might be better alternatives (based on technical merits) that take advantage of newer APIs in Svelte/Vite/DOM.

For `markdoc-svelte`, I can personally vouch for the package quality after using it in several projects and reviewing the source code. Also looking at the `m3-svelte` package, that one doesn't have that many more stars and downloads compared to `markdoc-svelte`: https://github.com/sveltejs/svelte.dev/pull/1592/files#diff-8398f3534bbf7138cc626b60d490018629ddde7d2044c8a6e387f5161a6edd1eR8-R9

Just like it's valuable to showcase a Material Design component library alongside other component libraries, adding a Markdoc package is a valuable addition to show Svelte has good support for more than Markdown.

With that said, I hope you will reconsider merging this. If not, that's unfortunate but I guess we'll just have to do some marketing. And a disclaimer - I'm not the original author of `markdoc-svelte`, just a happy user and small contributor :)

---

## Proposal for handling similar cases in the future
What if the `svelte.dev` packages also included smaller projects as long as they meet a high quality? Perhaps a dedicated section for up-and-coming packages that is clearly labelled like "Explore up-and-coming packages". As long as packages meet a high quality and at least a few people in the community support it by for example voting and commenting on a GitHub issue, the package would be included in the "up-and-coming" section.

This could use a GitHub issue template in the `svelte.dev` repo, and a dedicated issue label to make it easy for the community to suggest new packages and discuss why the package should be added. We also need clear rules for packages which packages are included, so we have a common understanding of when to add packages - and when to remove them due to inactivity or similar reasons.

If a wider range of Svelte ecosystem projects get visibility and usage, they will benefit from feedback and development, which ultimately will help the Svelte ecosystem grow and improve faster.

If we don't deal with this dynamic, this situation would risk becoming a prime example of the systems trap `Success to the successful - Competitive Exclusion`, as described on page 126 of the book [Thinking in Systems](https://en.wikipedia.org/wiki/Thinking_In_Systems%3A_A_Primer). The language of the screenshot-quote below is a bit extreme in this context, but I hope you get the point. We will all benefit from a diverse and thriving ecosystem - and diversity needs active support to not just point everyone to the most popular packages when there might be alternative solutions out there. Better solutions based on technical merits and new perspectives, rather than what is already popular.

image

This should probably be a separate issue.

teemingc:
@Greenheart I was revisiting this and wanted to suggest submitting `markdoc-svelte` to https://sveltesociety.dev/library . Anyone can make submissions and I think it would be really helpful to grow the collection of libraries there.

PR #1583, opened 2025-10-03
fix: Changes wording from stop to pause

Tellev:
As the word 'stop' is usually connected with stopping playback and resetting the playhead, I think the function we implement here should be called `pauseAll`. The implemented function allows resuming playback from where it was _paused_ and it even calls a function named `pause()`. Yes, I know this is nitpicky ... but I'm an audio guy 😅


Rich-Harris (approved):
thanks!

PR #1592, opened 2025-10-06
add m3-svelte to packages page

benmccann:


teemingc:
I tried some of the components on mobile and it doesn't work so well there. Makes me wonder if we need some min requirements for these component libraries such as:

- [ ] Works well with touch 
controls/mobile
- [ ] Accessible, can be keyboard controlled, passes screen reader checks, etc.

The community could also be involved in helping us test and fix up these issues. It'll also probably serve as good feedback to the authors.

Rich-Harris:
I gave it a spin just now on mobile and it seemed to work pretty well for the most part — perhaps some stuff has been fixed?

teemingc:
Yeah seems to be working fine now. I might have been confused by the icons which are circles and thought they are checkboxes

Rich-Harris:
cool, will go ahead and add it

PR #1597, opened 2025-10-07
fix: bug in reactive classes tutorial (app b)

M0ST4FA:
Issue #443

Make sure the embiggen button never embiggens rectangle dimensions beyond ```MAX_SIZE```.

Note: the same bug exists in "app a" but I don't understand why there are two apps in the first place, so I chose one randomly. Also, I checked the documentation repo and I couldn't find the tutorials in the directory tree, so sorry if the PR is in the wrong place.


Conduitry:
There are two apps because one is what the user is initially presented with and the other is the solved version.

This isn't the right fix, though. We handle `MAX_SIZE` in a later tutorial - https://svelte.dev/tutorial/svelte/getters-and-setters

I think, if anything, perhaps `MAX_SIZE` should be removed from the earlier tutorial.

PR #1623, opened 2025-10-18
Add black background color to App component

oliveiraa:
The default tutorial looks like this: 
image

This is a tiny change so the div has a background and its easier to see and understand where the mouse tracking would happen:
image


7nik:
1. On the white theme, it becomes black text on a black background.
2. It doesn't make sense, as the div has `width: 100%` and covers the entire App area. I'm not sure why you have a different result.

oliveiraa:
Sorry, my bad on the color choice for different themes. 

As for the not making sense part, while I agree with the theory, just using Chrome and you could see from the screenshot what it was rendering like.

With that said, I'll close it, thanks for checking it.

PR #1646, opened 2025-10-30
use remote functions

Rich-Harris:
WIP


apps/svelte.dev/src/routes/+layout.svelte
```diff
@@ -8,9 +8,10 @@
 	import { injectSpeedInsights } from '@vercel/speed-insights/sveltekit';
 	import { inject } from '@vercel/analytics';
 	import { beforeNavigate } from '$app/navigation';
+	import { get_banner, get_nav_links } from './data.remote.js';
 
-	injectSpeedInsights();
-	inject({ mode: dev ? 'development' : 'production' });
+	// injectSpeedInsights();
+	// inject({ mode: dev ? 'development' : 'production' });
```
dummdidumm: why is this commented out?

Rich-Harris: because it's annoying console spam. maybe we should put it inside a `if (!dev)` block?

PR #1406, opened 2025-07-04
fix: copy the correct code block

wyMinLwin:
fixes #1403

## Changes
`/packages/site-kit/src/lib/components/Text.svelte`

### Problem
This issue is due to selecting wrong targeted element.

### Solution
I changed query for selecting document.
```js
parent.querySelector(`pre[data-language="${ts ? 'ts' : 'js'}"] code`) as HTMLElement;
```

### Impact
The change improves the logic for selecting code blocks based on the language toggle by explicitly filtering for the `ts` or `js` language instead of relying on positional selectors.

PR #1647, opened 2025-10-30
fix: Correct copy for specific code block [Svelte]

wyMinLwin:
The problem is for **Svelte** code block which used **TypeScript** selecting the element which doesn't exist.

Code block selection improvements:

* Enhanced the query logic to check the code block's language and use `pre:last-of-type code` for **Svelte** samples when the **TypeScript** toggle is enabled, ensuring the correct code is copied for Svelte examples.
* Retained the original query for **TypeScript** and default cases, preserving expected behavior for other languages.


7nik:
I wonder if we can go with more simple
```js
const code = parent.querySelectorAll("code")[ts ? 1 : 0];
```

wyMinLwin:
> I wonder if we can go with more simple
> 
> ```js
> const code = parent.querySelectorAll("code")[ts ? 1 : 0];
> ```

No it's because if you do like that it would lead back to issue #1406 
which is copying wrong **TypeScript** code.

For current issue your solution will be fine but we need to think about what happened in issue #1406 as well

7nik:
`pre:last-of-type code` has the same problem, right? (any single `pre` is `:last` and `:last-of-type`)

We should just select only "top-level" code blocks: 
```js
const code = parent.querySelectorAll(":scope > pre > code")[ts ? 1 : 0];
```

wyMinLwin:
> `pre:last-of-type code` has the same problem, right? (any single `pre` is `:last` and `:last-of-type`)
> 
> We should just select only "top-level" code blocks:
> 
> ```js
> const code = parent.querySelectorAll(":scope > pre > code")[ts ? 1 : 0];
> ```


This work Great!
I updated the code. Thanks.

Issue #1576, opened 2025-10-02
`/packages` github/npm links/buttons are way too small


Hugos68:
<img width="184" height="93" alt="Image" src="https://github.com/user-attachments/assets/60295f4e-6709-475e-bb9c-f31c95c0574c" />

As you can see the size for these buttons are 14x14 both mobile and desktop, these are simply too small to accurately press on touch devices (my thumb alone is double that size!)

We should adhere to the WCAG spec for this: https://www.w3.org/WAI/WCAG22/Understanding/target-size-minimum.html

TL;DR: 24x24, preferably a little bit bigger than that as it's still small but willing to compromise.

Willing to create a PR 👍

Ocean-OS:
Closed by #1651

PR #1651, opened 2025-11-01
fix: size adjustment for GitHub and npm icons 

wyMinLwin:
# Adjust Mobile Icons size 

PR for #1576 
changed to **24x24** according to **WCAG spec**

In Mobile Size:

Screenshot 2025-11-01 at 18 34 49

In Desktop Size:

Screenshot 2025-11-01 at 18 33 04

### DIFF
--- a/apps/svelte.dev/src/routes/packages/PackageCard.svelte
+++ b/apps/svelte.dev/src/routes/packages/PackageCard.svelte
@@ -81,7 +81,7 @@
 	<!-- this is a sibling element so that we don't have links inside links -->
 	{#if !pkg.svAlias}
 		<div class="links">
-			<span style="display: flex;">
+			<span>
 				<a
 					href="https://npmjs.org/package/{pkg.name}"
 					target="_blank"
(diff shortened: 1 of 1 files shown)


Ocean-OS:
thank you!

apps/svelte.dev/src/routes/packages/PackageCard.svelte
```diff
@@ -81,7 +81,7 @@
 	<!-- this is a sibling element so that we don't have links inside links -->
 	{#if !pkg.svAlias}
 		<div class="links">
-			<span style="display: flex; gap: 0.75rem">
+			<span style="display: flex;">
```
7nik: I think `display: flex` can be moved to CSS as well. And I'm not sure there is a reason for the span.

wyMinLwin: I moved `display: flex` to CSS.

Span hold two icons as container and deal inner styling like gap between icons so I think it's ok to keep it.

Issue #696, opened 2024-10-26
Playground / REPL enhancements


techniq:
### Describe the problem

Based on a Twitter/X [discussion](https://x.com/techniq35/status/1849880234381677043), and at the [request](https://x.com/Rich_Harris/status/1850204736995176893) of @Rich-Harris, I would like to start a discussion on possibly adding some additional features/functionality to the playaround (aka REPL).  I thought opening a general issue to discuss features/scope, and then spiking off individual issues as we determine their usefulness and feasibility.

Initial ideas:
- Full page / presentation mode (with URL)
  - Remove top header
  - Show/hide code, preview, console, etc
- Screenshot / previews (similar to codepen, codesandbox, observable notebooks (🫗), etc)
- Ability to browse all of a user's REPLs, and/or the ability to "publish" / mark as public
  - Currently the security model is similar to "Unlisted" on YouTube where anyone can see a REPL if you have the link, and doesn't warrant itself to exploration
  - Going further with "new", "trending", etc would be nice, but likely well outside the initial scope.

### Describe the proposed solution

Discuss these (and other ideas) to determine their usefulness, feasibility, and possible design/architecture.

### Importance

nice to have

PR #1664, opened 2025-11-13
feat(repl): add font size adjustment to editor

ssssota:
Added a font size change feature to the REPL editor. This is necessary to make the code more readable when screen sharing or projecting the Playground for Svelte promotion.
(Since there was no discussion in the issue, it's okay to close it if it doesn't meet your intentions.)

image

### DIFF
--- a/packages/repl/src/lib/Editor/Editor.svelte
+++ b/packages/repl/src/lib/Editor/Editor.svelte
@@ -55,6 +55,7 @@
 <!-- svelte-ignore a11y_no_static_element_interactions -->
 <div
 	class="container"
+	style="--editor-font-size: {workspace.fontSize}px"
 	bind:this={container}
 	onpointerdown={() => {
 		workspace.enable_tab_indent();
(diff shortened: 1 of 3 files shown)


7nik:
The standard page zoom should work much better for this purpose: it scales everything, including the result section (and other tabs), docs, etc. How is this better?

ssssota:
Thank you for your prompt confirmation!

With page zoom, the header is also enlarged, which can cause the header to become too large when projected.
Given that general code editors are equipped with font size adjustment features in addition to zoom functionality, I believe there is a certain validity to enlarging only the code's font.

teemingc:
@ssssota Perhaps this can be converted to an issue for further discussion? I think the problem could also be solved by being able to hide the header for presentations 😄 some kind of "presentation mode" toggle might be nice for this purpose

ssssota:
@teemingc Yes, converting this to an issue sounds fine to me.

teemingc:
Awesome. Seems like what we've talked about is already mentioned in https://github.com/sveltejs/svelte.dev/issues/696

Issue #1657, opened 2025-11-09
Following the tutorial there are minor annoyances


mmaterowski:
### Describe the bug

1. After clicking `solve` for https://svelte.dev/tutorial/kit/the-form-element there is bunch of compiler warnings regarding unused scss.

<img width="490" height="300" alt="Image" src="https://github.com/user-attachments/assets/dec848c7-dde5-4f39-989e-51511e8c6939" />

2. When trying to create a file under  https://svelte.dev/tutorial/kit/pages I mistakenly enetered `*page.svelte` instead of `+page.svelte` and I got unclear warning about what my mistake is.   

<img width="788" height="383" alt="Image" src="https://github.com/user-attachments/assets/1d38ac97-f6e0-41e5-801e-ddb50f82ae0d" />

3. I have a dark theme set on my mac, and the tutorial page seems to reflect it. Once i add a ToDo in https://svelte.dev/tutorial/kit/form-validation the page changes to 'normal' mode and background is white instead of gray. 

Those are minor annoyance but could be definitely improved. I found the tutorial cool , but such small things make a tutorial look less professional than it is. Once i find more time during the week I will try to fix them myself.

### Reproduction

Reproduction steps above

### Logs

```Shell

```

### System Info

```Shell
MacOS 26.0.1 
No specific info, those infos are relevant to tutorial
```

### Severity

annoyance

### Additional Information

Tested on brave browser

GrygrFlzr:
For (1) it looks like it's not that the `solve` button creates the warnings, but rather a combination of two quirks:
1. Those are messages about the *pre-solved* state of the exercise that are present from the start
2. The editor does not reprocess after the `solve` button is pressed, so it just retains the pre-existing index ranges of where the previous warnings were, which results in the really weird rendering

teemingc:
3rd issue is related to https://github.com/sveltejs/svelte.dev/issues/1111

PR #1667, opened 2025-11-15
feat: add editing_constraints for layouts tutorial page

mmaterowski:
As mentioned in #1657 (2): After attempting to create a file with invalid filename the modal with unclear message pops up - editing_contraints were missing for Layouts file. 

After the fix the correct, expected route is shown when trying to create file with unwanted filename

- [ x ] It's really useful if your PR references an issue where it is discussed ahead of time.
- [ x ] Prefix your PR title with `feat:`, `fix:`, `chore:`, or `docs:`.
- [ x ] This message body should clearly illustrate what problems it solves.


Rich-Harris:
Thank you but fixing this manually one exercise at a time isn't sustainable — we need to do this work automatically. Opened #1813

PR #1707, opened 2025-12-14
docs: clarify timer accumulation in $effect tutorial

mdmahendri:

## What does this PR do?
- Clarifies the explanation about the interval timer example in the tutorial
- Explain in more detail that slow down doesn't work because multiple timers accumulate


Rich-Harris:
Thank you but this feels unnecessary — the text already explains what's happening. In fact the proposed change is _less_ clear: whereas at present it says that the effect runs 'each time `interval` gets smaller', the changed text connects the effect to the event (the button click) rather than the state change that happens as a _result_ of the event.

PR #1715, opened 2025-12-21
docs: Add proxy behavior example to Deep State tutorial

atmosuwiryo:
Adds a concrete code example to clarify that mutations to $state() proxies don't affect the original object:

```js
let original = [1, 2, 3, 4];
let numbers = $state(original);

numbers.push(5);  // numbers is now [1,2,3,4,5]
console.log(original); // Still [1,2,3,4] - unchanged!
```

Makes the existing NOTE section clearer for learners.


sacrosanctic:
Do we need an example for every statement on the docsite?

atmosuwiryo:
You're right that we don't need examples for everything. 

My concern was that the current code in the REPL `let numbers = $state([1, 2, 3, 4])` doesn't actually demonstrate the proxy behavior mentioned in the NOTE - there's no 'original object' to show the separation from. 

The NOTE felt disconnected from the REPL code.

Rich-Harris:
Thank you but I'm inclined to agree with https://github.com/sveltejs/svelte.dev/pull/1715#issuecomment-3684725339 — this feels like it adds a lot of unnecessary noise to the page.

Issue #1722, opened 2025-12-29
DOCS: Incorrect Highlight in https://svelte.dev/tutorial/svelte/component-events


emerson-proenca:
Small issue in https://svelte.dev/tutorial/svelte/component-events
<img width="1350" height="505" alt="Image" src="https://github.com/user-attachments/assets/187249db-c786-40b9-a0dd-1a988374fa20" />

The docs highlight parts of HTML that do not change, `>-1</` and `>+1</` these don't need Highlight. I might be wrong here, but I'm pretty sure just changing it to, would solve this issue.:
```svelte
/// file: Stepper.svelte
<button +++onclick={decrement}+++ >-1</button>
<button +++onclick={increment}+++ >+1</button>
```

Created a PR at #1723

Happy new year!

PR #1723, opened 2025-12-29
docs: Highligth in Basic/SvelteEventsComponent/events

emerson-proenca:
**Related:** Issue #1722

Small issue in https://svelte.dev/tutorial/svelte/component-events
Image

The docs highlight parts of HTML that do not change, `>-1</` and `>+1</` these don't need Highlight.

Happy new year!


Rich-Harris:
Thank you but this doesn't work, it just results in this instead which is even worse:

image

The logic needs a complete overhaul. Opened #1812

PR #1730, opened 2026-01-14
docs: enhance note on spread props for implicit props

ajitzero:
Emphasize the importance of destructuring props for nested components, particularly when implicit props such as children may be present.

When forgoing the destructuring, `children` is included in the value when there are child elements. Spreading these on a child component, which has its own children element, causes conflict and is hard to debug due to the implicit nature of this (magic).

I've added this note while debugging some issues with an SVG-based component, where nearly every prop needed to be forwarded, but also needs a lot of different child elements conditionally.

I acknowledge that this is probably a non-issue for experienced Svelte devs, but this stumped me for quite some time as someone primarily working with Angular & Vue. Please feel free to close this if this is redundant.

### DIFF
--- a/apps/svelte.dev/content/tutorial/01-svelte/03-props/03-spread-props/index.md
+++ b/apps/svelte.dev/content/tutorial/01-svelte/03-props/03-spread-props/index.md
@@ -40,3 +40,9 @@ We _could_ fix it by adding the prop...
 > ```js
 > console.log(stuff.name, stuff.version, stuff.description, stuff.website);
 > ```
+>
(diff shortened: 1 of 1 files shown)


sacrosanctic:
Doesn't seem like there are many reasons not to destructure. But I wouldn't add that note; it should instead be a eslint rule.

sacrosanctic:
On second thought, it is by design that everything can be forwarded to a child component. This is not as clear-cut.

@ajitzero, I think you should open an issue and provide reproductions and examples of your problem.

ajitzero:
> it should instead be a eslint rule.

This makes sense. My intention was to inform the existence of implicit props, such as children, since, until that point in the tutorial, props are just "inputs" and attributes. The current note on destructuring further solidifies the notion that props are just attributes, and newer devs may not handle this case.

I'll try to word this better with an issue later.

PR #1735, opened 2026-01-16
Store tailwind option when saving playground apps

### LINKED ISSUE #1220: Store `tailwind` option in database
When toggling the Tailwind option (#1218, #1219) we update the hash in the URL bar. We also need to store the option in the database when an app is saved

mustafa0x:
## Summary
- send tailwind flag when creating/saving playground apps
- persist/retrieve tailwind via gist create/update/read
- hydrate repl tailwind from saved gist in playground + embed

## Caveat
- Assumes DB schema/RPCs already support `gist.tailwind` and parameters on `gist_create`/`gist_update`; otherwise those DB changes are still needed.

Closes #1220

### DIFF
--- a/apps/svelte.dev/src/lib/db/gist.js
+++ b/apps/svelte.dev/src/lib/db/gist.js
@@ -38,13 +38,14 @@ export async function list(user, { offset, search }) {
 
 /**
  * @param {User} user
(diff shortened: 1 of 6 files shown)


Rich-Harris (approved):
thank you!

PR #1736, opened 2026-01-16
add security.txt

Rich-Harris:
Adds a `.well-known/security.txt` file per [RFC 9116](https://www.rfc-editor.org/rfc/rfc9116)

### DIFF
--- a/apps/svelte.dev/static/.well-known/security.txt
+++ b/apps/svelte.dev/static/.well-known/security.txt
@@ -1,2 +1,4 @@
 # Please report security issues via GitHub:
 Policy: https://github.com/sveltejs/.github/blob/main/SECURITY.md
+Contact: mailto:security@svelte.dev
+Expires: 2027-02-28T00:00:00z


teemingc:
Apparently there are two fields which must always be present:
- Contact https://www.rfc-editor.org/rfc/rfc9116#section-2.5.3
- Expires https://www.rfc-editor.org/rfc/rfc9116#name-expires

apps/svelte.dev/static/.well-known/security.txt
```diff
@@ -0,0 +1,2 @@
+# Please report security issues via GitHub:
+Policy: https://github.com/sveltejs/.github/blob/main/SECURITY.md
```
teemingc, line 2:
```suggestion
Policy: https://github.com/sveltejs/.github/blob/main/SECURITY.md
Contact: mailto:security@svelte.dev
Expires: 2027-02-28T00:00:00z
```
The `Expires` field is required and is recommended to be set to a year but that can be annoying to remember to update https://www.rfc-editor.org/rfc/rfc9116#section-2.5.5

PR #1752, opened 2026-01-23
chore: Allow cloning from repositories with different names from their upstream

elliott-with-the-longest-name-on-github:
This updates the sync-docs workflow so that it can pull from a downstream repo that has a different name from the upstream repo, a semi-common (and very confusing / hard to diagnose) failure mode for the sync-docs workflows. Previously, the CLI could be invoked as such:

```sh
sync-docs -p kit
sync-docs -p kit#branch-name
```

Now, it has only one accepted format:

```sh
sync-docs -p upstream-name#downstream-name#branch-name
```

Where `upstream-name` is one of the allowed repos (`kit`, `cli`, `svelte`, `mcp`) and `downstream-name` is anything. This also adjusts the workflows to use the new format (for backwards compatibility). Once I update the svelte docs bot, I'll remove the backwards-compatible parameters and update the new ones to be required.

### DIFF
--- a/.github/workflows/docs-preview-create.yml
+++ b/.github/workflows/docs-preview-create.yml
@@ -5,7 +5,20 @@ on:
     inputs:
       repo:
         description: 'Repository name (not fully-qualified, eg. `svelte` or `kit`)'
-        required: true
+        required: false
+        type: string
+      upstream:
+        description: 'Upstream repository (eg. `svelte` or `kit`)'
+        required: false
+        type: choice
+        options:
+          - svelte
+          - kit
+          - cli
+          - mcp
+      downstream:
+        description: 'Downstream repository (eg. `svelte` or `kit`)'
+        required: false
         type: string
       pr:
         description: 'PR number'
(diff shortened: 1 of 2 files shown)


Rich-Harris:
When would the repo names be different?

Could we use a different pattern and make the downstream name optional, so that e.g. `kit#name` and `kit:alias#name` work?

Conduitry:
The repo name would be different if the person forking the repo chooses to do so. For example, I use repo names like https://github.com/Conduitry/sveltejs_svelte for forks, to avoid collisions on common things like `docs` which I may have more than one forked repo called.

elliott-with-the-longest-name-on-github:
> When would the repo names be different?

What @Conduitry said. Anyone can name their fork anything they want.

> Could we use a different pattern and make the downstream name optional, so that e.g. kit#name and kit:alias#name work?

I did consider this but erred on the side of "just be explicit", happy to change it if you feel strongly about it. "Just remember this format" seemed better than "ugh, which delimiter am I supposed to use here again?"

Conduitry:
This script isn't something that anyone is ever going to be calling manually, is it? If it's only going to be called by other actions, I would also argue for simple and explicit, even if verbose.

elliott-with-the-longest-name-on-github:
> This script isn't something that anyone is ever going to be calling manually, is it? If it's only going to be called by other actions, I would also argue for simple and explicit, even if verbose.

AFAIK this method of usage (explicit positional arguments) is only used by CI. If you're using it locally you should be symlinking your other repos to `repos/{kit|cli|mcp|svelte}` which means you don't have to specify any of this

Rich-Harris:
Ah fair enough. In that case keeping it explicit/simple seems reasonable. Would personally still incline towards `up:down#branch` as that seems more obvious to me but also I couldn't care less as long as it works

PR #1764, opened 2026-01-30
fix: consider `~~~` a codeblock

paoloricciuti:
In the MCP docs, we need to show a codeblock for markdown with codeblocks in them. This can get really messy but luckily you can use `~~~` instead of triple backtick to do exactly that...however it was not considered "in a codeblock" meaning it was generating the table of content for them (and failing to deploy since it was a link prerendering was visiting that was not actually there).

This fixes it.


paoloricciuti:
Not needed anymore

PR #1768, opened 2026-02-03
fix: decode html entities before passing jsdoc to renderer

bsmithgall:
On the current doc website, HTML content described in JSDoc comments are not unescaped before being passed along to the markdown renderer. When Shiki extracts the JSDocs from the other repos, it escapes the HTML (e.g. `<` becomes `&#x3C;`). This is then forwarded along directly to the markdown renderer. 

This change manually reverses this process for a handful of characters. See attached screenshots for a before and after.

Before:
before

After:
after


Ocean-OS:
Thank you!

PR #1720, opened 2025-12-24
feat(doc): add reference link

jycouet:
Closes: #644

- With `reference` as a link to the right place in the doc
image

---

Closes: #1283

- fixing escaping tooltips & no highlight in tooltips

<table>
<tr>
 <td>Before</td>
 <td>After</td>
</tr>
<tr>
 <td>image</td>
 <td>image</td>
</tr>
</table>

### Before submitting the PR, please make sure you do the following

- [x] It's really useful if your PR references an issue where it is discussed ahead of time.
- [x] Prefix your PR title with `feat:`, `fix:`, `chore:`, or `docs:`.
- [x] This message body should clearly illustrate what problems it solves.

Issue #747, opened 2024-11-01
Link digests between tutorial, guides, and types; and feature summaries


theetrain:
I've gathered some site feedback from my coworkers and I feel some of them are worth sharing as a single ticket since they impact the reading experience when planned together. If you prefer I split these into separate tickets let me know.

## Have tutorial pages link to respective docs

At the bottom of a tutorial page, add a 'read more' section that points to one or more relevant guides in the docs.

## Have docs link to respective tutorials

Near the top of a guide or page, include a link to the respective tutorial page to help those get started hands on.

## Have guides link to types

On a given guide, wherever types are mentioned, have them link to the respective type.

Example, these could be links to their types:

![image](https://github.com/user-attachments/assets/7e87a06c-b0e9-4354-8466-c1bfc811376b)

## Provide a feature summary per guide

Some guides could benefit from a `<table>` or feature summary. For example when referencing form actions, it takes a bit of work to know the difference between `use:enhance`, `applyAction`, and `update`. A table could help as an addition to the explanation of those features, like so:

When supplying a function to `use:enhance`, its callback function behaviours are:

| effect | default (no method) | `update` | `applyAction`
| --- | --- | --- | --- |
| reset form | no | yes | no |
| call `invalidateAll` | no | yes | no |
| call `goto` | no | yes | only when `result.type` === 'redirect' |

Also related: https://github.com/sveltejs/kit/issues/11823

## Provide a types index per guide

Somewhere on a given page, such as 'loading data' or 'form actions', provide a link farm or index that lists all related types so they can be looked up easily.

## Optionally display inherited types

On a given types reference, display all inherited types or optionally hide them behind an accordion. For example, `ServerLoadEvent` extends `RequestEvent` and it would be useful to see the provided features of `RequestEvent` such as `fetch` or `setHeaders` while browsing `ServerLoadEvent`. 

What's often the case is we're in VSCode previewing the event type for `+page.server.js@load` and we then look up the type for `ServerLoadEvent` only to hunt for all its types and features. Similarly, if a reader can jump from the "Loading Data" guide to `ServerLoadEvent`, then it may take some adjustment (or readjustment on a subsequent visit) to learn about its inherited properties.

PR #1775, opened 2026-02-06
feat: crosslinks

Rich-Harris:
Draft for discussion. A request we've had a few times is to have links between e.g. reference docs and the tutorial. This adds a mechanism for generating those crosslinks programmatically using tags, as well as a `<RelatedLinks>` component.

For example https://svelte.dev/docs/svelte/transition would look like this...

---

image

---

...while the tutorial page would look like this:

---

image

---

Some of the tags would need to be updated in this repo, but docs tags would need to be added in the respective repos. Right now I just hardcoded some example `transition` tags. Later we could expand this mechanism to blog posts and examples if we wanted.

### DIFF
--- a/apps/svelte.dev/content/tutorial/01-svelte/01-introduction/04-styling/index.md
+++ b/apps/svelte.dev/content/tutorial/01-svelte/01-introduction/04-styling/index.md
@@ -1,5 +1,6 @@
 ---
 title: Styling
+tags: styles-scoped
 ---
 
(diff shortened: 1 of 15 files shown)


benmccann:
https://github.com/sveltejs/svelte.dev/pull/1720 is also related. It generates some automatically that we used to have

theetrain:
This looks great! It would help solve most of https://github.com/sveltejs/svelte.dev/issues/747.

PR #1812, opened 2026-02-24
Fix highlighting

### LINKED ISSUE #1722: DOCS: Incorrect Highlight in https://svelte.dev/tutorial/svelte/component-events

Rich-Harris:
this results in way better inline diff highlighting. closes #1722 and probably others

## before

image

image

image

image

image

## after

image

image

image

image

image

In an ideal world we'd have unit tests around this stuff, but... we don't

### DIFF
--- a/packages/site-kit/src/lib/markdown/renderer.ts
+++ b/packages/site-kit/src/lib/markdown/renderer.ts
@@ -863,24 +863,36 @@ const delimiter_patterns = Object.fromEntries(
 	])
 );
 
-function highlight_spans(content: string, classname: string) {
-	const open_count = content.split('<span').length - 1;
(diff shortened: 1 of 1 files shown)


teemingc (approved):

jycouet:
Love it !

PR #1815, opened 2026-02-24
fix: correctly slugify special section titles

teemingc:
Searching for "details bind:open" in the site gives you the wrong link, the section title "`<audio>`" can't be searched for, italicised words show the markdown syntax in the search results, etc.

This PR fixes the slugify function for special titles such as "`bind:_property_`", etc. so that they show up correctly in the search results _and_ link to the correct section.

### DIFF
--- a/packages/site-kit/src/lib/types.d.ts
+++ b/packages/site-kit/src/lib/types.d.ts
@@ -21,6 +21,7 @@ export interface Document {
 	file: string;
 	metadata: {
 		title: string;
+		/** Prevents embedded examples from being displayed in the search */
 		private?: boolean;
 		[key: string]: any;
 	};


apps/svelte.dev/src/lib/server/content.ts
```diff
@@ -19,6 +19,20 @@ const assets = import.meta.glob<string>(['./**/+assets/**', './**/+assets/**/.en
 	base: '../../../content'
 });
 
+// the embed examples are not meant to show up in the search results
```
teemingc: Currently, you can search for "Untitled" and get this search result.

untitled search result

It's better if these "invisible" examples are removed from our search

packages/site-kit/src/lib/server/content/index.ts
```diff
@@ -32,21 +32,21 @@ export async function create_index(
 
 		const sections = Array.from(body.matchAll(/^#{2,3}\s+(.*)$/gm)).reduce((arr, match) => {
 			if (is_in_code_block(body, match.index || 0)) return arr;
-			const title = smart_quotes(match[1])
+			const title = match[1];
```
teemingc, line 35: Previously, we were slugifying the displayed title which has html in it. By differentiating the actual title and the displayed title, we can get the correct slug result

apps/svelte.dev/src/lib/server/content.ts
```diff
@@ -19,6 +19,20 @@ const assets = import.meta.glob<string>(['./**/+assets/**', './**/+assets/**/.en
...
+	if (path.startsWith('./examples/99-embeds')) {
+		delete documents[path];
+	}
+}
+
+for (const path of Object.keys(assets)) {
+	if (path.startsWith('./examples/99-embeds')) {
+		delete assets[path];
+	}
+}
```
Rich-Harris: we shouldn't delete these, because then the embeds vanish from e.g. https://svelte-dev-git-fix-search-for-special-titles-svelte.vercel.app/blog/svelte-css-in-js. We should just prevent them from being indexed (maybe with some frontmatter like `private: true` or `indexed: false` or whatever)

teemingc: Thanks for catching that! I've made the change and tested it

PR #1823, opened 2026-02-27
docs: rename mcp to ai

paoloricciuti:
Complete the rename of `sveltejs/mcp` to `sveltejs/ai-tools`


.github/workflows/docs-preview-create.yml
```diff
@@ -11,7 +11,7 @@ on:
           - svelte
           - kit
           - cli
-          - mcp
+          - ai
```
Rich-Harris, line 14: is this right? should it be `ai-tools`?

paoloricciuti: I had the same doubt but I think it should be `ai` since is what is used for the link and then it uses the repo property which I've set to `ai-tools`...but we might need to update the bot

Rich-Harris:
it still says 'MCP' in the dropdown, do we want to change that to 'AI'?

image

The `/docs/ai/overview` page is very MCP-centric, don't know if we want to change that to be a more general overview with a new `/docs/ai/mcp` section or if you'd rather do that in a follow-up

Rich-Harris:
well you can ignore my previous comment on two counts: a) it's outdated and b) the dropdown apparently disappeared when i took the screenshot?

paoloricciuti:
Let's do it in a follow up

PR #1837, opened 2026-03-04
feat: async mode toggle

Rich-Harris:
makes async/non-async discrepancies easier to repro

### DIFF
--- a/packages/repl/src/lib/Repl.svelte
+++ b/packages/repl/src/lib/Repl.svelte
@@ -144,6 +144,7 @@
 				svelte_version: svelteVersion,
 				onversion: (version, supports_async) => {
 					workspace.set_svelte_version(version);
+					workspace.supports_async = supports_async;
 					if (!supports_async && workspace.compiler_options.async) {
 						workspace.update_compiler_options({ async: false });
 					}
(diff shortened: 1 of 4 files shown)


packages/repl/src/lib/Bundler.svelte.ts
```diff
@@ -8,6 +8,7 @@ export default class Bundler {
 	#worker: Worker;
 
 	result = $state.raw<BundleResult | null>(null);
+	supports_async = $state(false);
```
dummdidumm: looks unused, only assigned to

packages/repl/src/lib/Repl.svelte
```diff
@@ -141,8 +142,11 @@
 		? new Bundler({
 				// svelte-ignore state_referenced_locally
 				svelte_version: svelteVersion,
-				onversion: (version) => {
+				onversion: (version, supports_async) => {
 					workspace.set_svelte_version(version);
+					if (!supports_async) {
+						workspace.update_compiler_options({ async: false });
```
dummdidumm, line 149: this could result in compiler options firing unnecessarily when async is already false. Either we compare against current option here or have a check in update_compiler_options

packages/repl/src/lib/Bundler.svelte.ts
Rich-Harris: good catch, fixed

packages/repl/src/lib/Repl.svelte
Rich-Harris: fixed

PR #1839, opened 2026-03-04
add an AGENTS.md

Rich-Harris:
lots more we could add here, figured we'd start small

### DIFF
--- a/.github/copilot-instructions.md
+++ b/.github/copilot-instructions.md
@@ -1,0 +1,1 @@
+AGENTS.md
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,1 @@
+AGENTS.md


elliott-with-the-longest-name-on-github (approved):

MathiasWP:
You can also instruct Claude to do this via [hooks](https://code.claude.com/docs/en/hooks-guide), or better IMO, add this as a pre-commit git hook that runs prettier only on staged filed. It’s really fast, and then it won’t be possible to push unformatted code anymore. Works great for both agents and humans. I would also suggest running lint on staged files as well in the pre-commit hook!

elliott-with-the-longest-name-on-github:
Claude-specific stuff is a no-go, a lot of us use Copilot / Opencode which don't implement the same features. A precommit hook might be a good idea, though. The point of this PR wasn't so much to make sure prettier gets run, though, as it was to stub out the AGENTS.md file and the other symlinked ones.

PR #1849, opened 2026-03-07
chore: additional explanation if prerender fails

teemingc:
@Rich-Harris if we update a type definition without changing a documentation page that uses it in a code block, twoslash will reuse a cached version of the code block that references an id which may no longer exist. This will cause the prerendering to fail with "this page referenced an ID that does not exist".

This PR adds remediation steps if the prerendering ever fails and that is the case

EDIT: In hindsight, I wonder if this means there are pages which do not update their codeblock rendered "types on hover" when we update their types. Maybe we need to ditch caching entirely or just make sure to invalidate the cache when deploying on main


Rich-Harris:
Ha, there's a reason this is one of computer science's [two hard things](https://martinfowler.com/bliki/TwoHardThings.html) https://github.com/sveltejs/svelte.dev/pull/1841#pullrequestreview-3891011572

Issue #1177, opened 2025-03-05
The examples in the type info tooltips are a single line


teemingc:
The examples shown in the type hover tooltips (`@example`) are a single line. It would be more readable if they were rendered as a block of code.

![example on a single line](https://github.com/user-attachments/assets/4f55a898-c581-4bab-8bc4-b4187171237c)

PR #1864, opened 2026-03-16
fix: render JSDoc @example as block code in tooltips (#1177)

moaaz-ae:
Enables multi-line rendering and syntax highlighting for `@example` tags in documentation tooltips by using the full markdown renderer.

Resolves:

- https://github.com/sveltejs/svelte.dev/issues/1177

### Before

1177-before

### After

1177-after


teemingc:
This is brilliant. Thank you. I’m letting the Vercel preview build run

PR #1875, opened 2026-03-18
fix: missing .env file in environment variables lessons (#1835)

### LINKED ISSUE #1835: .env file is not available in Environment variables lessons
.env file is supposed to be editable in four lessons of [Advanced SvelteKit / Environment variables](https://svelte.dev/tutorial/kit/env-static-private) section of Tutorial. In fact this file is not presented in the file system of mentioned lessons at all.

Image

AbdullahShafqat-OG:
No `.env` files were showing up in the Environment Variables lessons of the Advanced SvelteKit tutorial. This PR updates the glob import so that `.env` is correctly included in the build version and visible across all four lessons in this section.

Fixes [#1835](https://github.com/sveltejs/svelte.dev/issues/1835)

### Before
image

### After
image  

**_Note:_** 
_Verified that the `__data.json` for `env-static-private` route includes the .env file in build_
image

### DIFF
--- a/apps/svelte.dev/src/lib/server/content.ts
+++ b/apps/svelte.dev/src/lib/server/content.ts
@@ -13,6 +13,7 @@ const documents = import.meta.glob<string>('./**/*.md', {
 });
 
 const assets = import.meta.glob<string>(['./**/+assets/**', './**/+assets/**/.env'], {
+	exhaustive: true,
 	eager: true,
 	query: '?url',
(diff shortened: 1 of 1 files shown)


teemingc (approved):
Thanks! I had to adjust the patterns to ignore `node_modules` and `.svelte-kit` since adding `exhausted: true` included them. Other than that, it's a good change

PR #1882, opened 2026-03-21
feat: build cli types during docs sync

jycouet:
Adds `@sveltejs/sv-utils` types support for twoslash hover info in cli docs.

- Post-clone hook runs `pnpm install && pnpm build` on the cli repo so `.d.mts` types are generated
- Symlinks `sv` and `@sveltejs/sv-utils` in `node_modules` to the cloned repo so twoslash resolves PR branch types
- Adds `@sveltejs/sv-utils` to twoslash compiler types

Related: https://github.com/sveltejs/cli/pull/997

### DIFF
--- a/apps/svelte.dev/scripts/sync-docs/index.ts
+++ b/apps/svelte.dev/scripts/sync-docs/index.ts
@@ -24,7 +24,6 @@ interface Package {
 	docs: string;
 	types: string | null;
 	process_modules?: (modules: Modules, pkg: Package) => Promise<Modules>;
-	/** Runs after cloning the repo (e.g. to install deps and build types) */
 	post_clone?: (dir: string) => Promise<void>;
 }
 


Rich-Harris:
I wonder if we should generalise this, so that we use the PR for Svelte and Kit types as well? This is a frequent nuisance on those repos

jycouet:
> I wonder if we should generalise this, so that we use the PR for Svelte and Kit types as well? This is a frequent nuisance on those repos

Good idea! Done.
Added `post_clone` for svelte and kit that patches `node_modules/<pkg>/types/` with the cloned repo's types dir. Since both repos commit `types/index.d.ts` -> no build is needed. Just copying the types/ folder is enough for twoslash to resolve mports against the PR branch. right?

WDYT?

Rich-Harris:
sounds great! let's try it

teemingc:
Thank you! Been thinking about how to do this and thought about making CI run pnpm i with overrides on workspace repos. This is better

PR #1894, opened 2026-03-25
fix: prevent overlay flash when closing iOS mobile nav menu (#1548)

### LINKED ISSUE #1548: Flicker when closing the mobile nav menu
The overlay flashes whenever the mobile nav menu is closed. Tested on iPhone 16 iOS 26

https://github.com/user-attachments/assets/d856e910-90d9-4f0f-852d-7df8f6533380

AbdullahShafqat-OG:
Fixes a visual glitch where the overlay briefly flashes when closing the mobile navigation menu.

The issue was caused by a discrepancy in transition durations between `MobileMenu` and `ModalOverlay`. 

The transition timing has been streamlined by defining it in `MobileMenu` and passing it down as a prop to `ModalOverlay`. This ensures both components stay in sync and eliminates the visual inconsistency.

Fixes [#1548](https://github.com/sveltejs/svelte.dev/issues/1548)

### Before
https://github.com/user-attachments/assets/03a0b451-6105-4ef6-8edd-9d9ae356cf4c

###  After
https://github.com/user-attachments/assets/75de89d5-9bc5-44c8-9489-c2227fcb4558

### DIFF
--- a/packages/site-kit/src/lib/components/ModalOverlay.svelte
+++ b/packages/site-kit/src/lib/components/ModalOverlay.svelte
@@ -1,11 +1,11 @@
 <script lang="ts">
 	import { fade } from 'svelte/transition';
 
-	let { onclose }: { onclose?: () => void } = $props();
+	let { onclose, duration = 200 }: { onclose?: () => void; duration?: number } = $props();
 </script>
 
 <div
-	transition:fade={{ duration: 100 }}
+	transition:fade={{ duration }}
 	class="modal-overlay"
 	aria-hidden="true"
 	onclick={onclose}
--- a/packages/site-kit/src/lib/nav/MobileMenu.svelte
+++ b/packages/site-kit/src/lib/nav/MobileMenu.svelte
@@ -10,6 +10,8 @@
 	import type { NavigationLink } from '../types';
 	import ModalOverlay from '../components/ModalOverlay.svelte';
 
+	let transition_duration = 200;
+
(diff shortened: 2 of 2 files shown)


teemingc:
Sorry for taking a long time to look at this. I've tested on the latest iOS and this doesn't seem to be an issue anymore

AbdullahShafqat-OG:
@teemingc I just tried with iOS 26.0.1 on iPhone 15 and still flashes for me. Perhaps it does not show up in a later version of iOS. Should it still be addressed perhaps to fix in previous iOS versions as well?

teemingc:
You’re right. I’m wondering if we can preserve the quick transition on menu close though. Is the flash caused by the 100ms transition duration?

AbdullahShafqat-OG:
The menu animation in the deployed version is currently using 200ms already as well. The 100ms was for the modal-overlay which was the partially faded black panel on top of the screen and behind the navigation menu. 

Root cause of the flash: the overlay and menu had different durations/easings, so the overlay reached opacity: 0 before the menu finished. While it sat there waiting, iOS dropped the fill: forwards, reverting it to its pre-animation opacity hence the flash.


It does work for transition duration 100ms but that gets a very snappy feel and feels off from the original feel of the menu that was already there. Could add something that caters to this iOS caveat but I think we can keep the fix simple here also aligning the transition times for both the modal and the menu animation does make sense as well

PR #1923, opened 2026-04-09
fix(og): decode xml tags in og images

WarningImHack3r:
Fix OG images containing `<`s (among others) displaying text with escaped characters (`&lt;`).

This fix is taken directly (and adapted) from the code I wrote in Svelte Changelog ([here](https://github.com/WarningImHack3r/svelte-changelog/commit/96651dc63fa30631ffa82c291ae06aa6c2728914) then [here](https://github.com/WarningImHack3r/svelte-changelog/commit/9cacf25c9368242868f1dba0fc3d58d79a02b572)), as we use the same core logic and I encountered the exact same issue. :D

| image | image |
| :---: | :---: |
| Before ([svelte.dev](https://svelte.dev/docs/svelte/svelte-window/card.png)) | After (localhost) |

#### Notes on the code

- I took advantage of the existing `escape.js` file to add in my (opposite) logic, which is exactly made for this purpose (lucky me!). In the existing code, I lifted the regex out of the function so it can be computed only once, and made it dynamic so that everything derives from the single existing record. I also added more characters to the record. Finally, I added a null-coalescing check to the existing function to avoid any problem, even though it'll very likely never be used
- I duplicated the vnode logic to both the OG image generators, as a bunch of the logic is already duplicated between both, and I don't really know where to put it anyway; it's very dependent on the very computing code in the core of the generators. Let me know if you have any suggestions!

> [!NOTE]
> ~~The CI lint step fails because of the same problem existing on `master`, I can't do much about it :(~~ (outdated)


teemingc (approved):
Thank you! This is great

PR #1927, opened 2026-04-10
fix: frontmatter rendering as removal

paoloricciuti:
In https://github.com/sveltejs/ai-tools/pull/188 I've included the subagent definition which is a markdown codeblock with the frontmatter.

The current rendering strategy detect that as a removal because of 
```
---
[content]
---
```
and you can see how bad it looks
image

This fixes while still leaving the ability to do diffs inside markdown blocks (it just exclude the actual frontmatter) in md and yml codeblocks.


jycouet:
- ci is failing because since `svelte@5.55.1` it's pulling `esrap@2.2.4` and that brings this type issue. I pushed an override to see. https://github.com/sveltejs/esrap/pull/108 would help
- now: _(but I have to go right now)_ 
```
apps/svelte.dev check: Error: Type '() => ReturnType<import("svelte").Snippet>' is not assignable to type 'Snippet<[]>'.
```
- Other than that, the real change you want looks good ^^

dummdidumm (approved):

PR #1932, opened 2026-04-12
fix:added &amp; before &lt and &gt 

cxnexion:
So in tutorial it took for me a few minutes to realize that I need to use &amp;&lt; and &amp;&gt; instead of < and >. This change will show new developers the right way without guessing.

- [  ] It's really useful if your PR references an issue where it is discussed ahead of time.
- [yes ] Prefix your PR title with `feat:`, `fix:`, `chore:`, or `docs:`.
- [ yes] This message body should clearly illustrate what problems it solves.


cxnexion:
oh wait sorry im wrong a bit

PR #1959, opened 2026-04-30
fix: get twoslash to work properly with JSDoc

Rich-Harris:
Various fixes for type hovers in JSDoc snippets — turns this...

image

...into this, by providing twoslash with the correct context in the form of a filename (so it doesn't mistakenly think it's still dealing with an `ambient.d.ts` file) and by stripping `@type` tags:

image

It also turns this...

image

...into this, by using CSS grids properly:

image


teemingc:
Looks like there are some snippet type errors possibly due to the new format

Rich-Harris:
Gradually working through the newly-exposed type errors. They'll need to be backported to the other repos, but it's easier to begin by making the changes here

Rich-Harris:
Extracted the necessary upstream changes: https://github.com/sveltejs/kit/pull/15781

Since we can't merge Kit docs just yet (#1955 is blocked), no harm in merging these temporary changes

PR #1967, opened 2026-05-09
fix: Svelte tutorial warning removals and corrections

michael-255:
The Svelte team has done a great job with both the tutorials and the entire docs site!

I took note of a few things to fix while working through them. There isn't a specific issue for any of these changes that I could find.

- [X] Prefix your PR title with `feat:`, `fix:`, `chore:`, or `docs:`.
- [X] This message body should clearly illustrate what problems it solves.

### Changes

Added minimum needed attributes to remove the aria warnings on `AudioPlayer.svelte` in several tutorials.

**Advanced Bindings, Dimensions:**
Allow input box to be clickable if user deletes all text. This one caught me briefly.

**Svelte Element:**
Added a TODO comment to make it clear where the user should put the `svelte:element` block.
Using `&amp` in the instruction example so the user doesn't attempt to use `<{selected}>` directly which will cause this error:

```
Error compiling App.svelte

Expected a valid element or component name. Components must have a valid variable name or dot notation expression
https://svelte.dev/e/tag_invalid_name
```


apps/svelte.dev/content/tutorial/02-advanced-svelte/07-special-elements/06-svelte-element/+assets/app-a/src/lib/App.svelte
```diff
@@ -9,8 +9,9 @@
 	{/each}
 </select>
 
+<!-- TODO replace with svelte:element block -->
 {#if selected === 'h1'}
 	<h1>I'm a <code>&lt;h1&gt;</code> element</h1>
 {:else}
-	<p>TODO others</p>
```
teemingc, line 15: This was intentional. It's to show that it wouldn't make sense to manually implement each one

teemingc:
Thanks! Do you mind splitting the different fixes into separate PRs?

teemingc (changes requested):

apps/svelte.dev/content/tutorial/02-advanced-svelte/04-advanced-bindings/03-media-elements/+assets/app-a/src/lib/AudioPlayer.svelte
```diff
@@ -31,6 +31,9 @@
 			<span>{format(time)}</span>
 			<div
 				class="slider"
+				role="slider"
+				aria-valuenow={time}
+				tabindex="0"
```
teemingc, line 36: This gets rid of the warning but tabbing to it doesn't actually do anything (that I know of). Maybe it's because we don't have keyboard controls but I wonder if that might blow up the scope of the tutorial

PR #1973, opened 2026-05-14
docs: mention Safari keyboard navigation in attachments tutorial

### LINKED ISSUE #971: tutorial/svelte/actions does not work with default Safari settings
Safari handles page element tabbing in a weird way, which breaks [the tutorial about trapping focus with Actions](https://svelte.dev/tutorial/svelte/actions). _This is an issue with Safari_ — but the tutorial should **warn unsuspecting Safari users** that the step won't work as written.

One work-around for Safari users is to chord option-Tab instead. Another is to change Safari's global accessibility settings.

See here for background on how Safari's default settings affect normal tabbing behaviours:

- [Cannot focus button using tab key navigation on Safari - Stack Overflow](https://stackoverflow.com/questions/67901731/cannot-focus-button-using-tab-key-navigation-on-safari)
- [Safari ignoring tabindex - Stack Overflow](https://stackoverflow.com/questions/1848390/safari-ignoring-tabindex)

----

This issue is not related directly to #700 affecting the same page, but it's worth noting that Safari is affected by that bug as well. Together they make this page of the tutorial feel extra-janky on Safari.

NlDev-hub:
Fixes #971.

This adds a short note to the attachments tutorial explaining that Safari does not focus page controls with the Tab key by default. The tutorial asks readers to test focus trapping with Tab, so Safari users need to use Option+Tab or enable keyboard navigation for the exercise to behave as described.

Checks:
- `git diff --check`
- `npx --yes prettier --check apps/svelte.dev/content/tutorial/01-svelte/08-attachments/01-attach/index.md`


apps/svelte.dev/content/tutorial/01-svelte/08-attachments/01-attach/index.md
```diff
@@ -12,6 +12,8 @@ Attachments are essentially element-level lifecycle functions. They're useful fo
 
 In this app, you can scribble on the `<canvas>`, and change colours and brush size via the menu. But if you open the menu and cycle through the options with the Tab key, you'll soon find that the focus isn't _trapped_ inside the modal.
 
+> [!NOTE] Safari does not focus page controls with the Tab key by default. To follow this exercise in Safari, use Option+Tab or enable keyboard navigation in Safari's Advanced settings.
```
teemingc:
```suggestion
> [!NOTE] Safari only focuses text fields and pop-up menus with the Tab key by default. To follow this exercise in Safari, use Option + Tab or enable Tab to highlight in Safari's advanced settings.
```

teemingc (approved):

PR #1997, opened 2026-05-25
chore: update @types/node to v22.19.19

Botato300:
The code was using the `fs.globSync()` function ([added in Node.js v22](https://nodejs.org/docs/latest/api/fs.html#fsglobsyncpattern-options)), but the @types/node package installed in the project was version v20. As a result, the editor/IDE or another tool was reporting an error stating that the function did not exist.
So, simply update @types/node to version v22.


teemingc (approved):
Thank you!

PR #1999, opened 2026-05-26
chore: normalize path separators for Windows compatibility

Botato300:
When I ran the command `pnpm run sync-docs`, it returned the following error on Windows:
> Error: Not handled content\tutorial\01-svelte\09-transitions\08-key-blocks\index.md
    at get_path (file:///C:/Users/botato300/Desktop/svelte.dev/apps/svelte.dev/scripts/sync-docs/crosslinks.ts:35:11)
    at generate_crosslinks (file:///C:/Users/botato300/Desktop/svelte.dev/apps/svelte.dev/scripts/sync-docs/crosslinks.ts:66:16)
    at file:///C:/Users/botato300/Desktop/svelte.dev/apps/svelte.dev/scripts/sync-docs/index.ts:289:1

The code expected the path to use the **/** separator. But in Windows, the path used the **\\** separator, so none of the “case” statements matched, and an exception was thrown.
https://github.com/sveltejs/svelte.dev/blob/281039cca89bde455c5db72c75e1084e2bb9aaa2/apps/svelte.dev/scripts/sync-docs/crosslinks.ts#L26-L38

The solution was simply to normalize the path so that it uses **/** as a separator.


apps/svelte.dev/scripts/sync-docs/crosslinks.ts
teemingc, line 27: Can we try importing `node:path` and split with `path.sep` instead?

teemingc (changes requested):
See above

apps/svelte.dev/scripts/sync-docs/crosslinks.ts
Botato300: Yes, it's possible and it works! BUT I preferred to normalize the path so that you can always handle it using the / separator afterward.

With the path.sep approach, you'd have to use it every time you call split() or perform a similar operation. By normalizing the path instead, you can forget about using path.sep, and it will still work correctly. Which approach do you prefer?

teemingc: Mm you have a point. Let's try it.

teemingc (approved):
Thanks! I only renamed the variable name to snake case since that's what we use for non-public APIs

PR #2008, opened 2026-05-28
fix: improve tutorial terminal text contast in dark mode

### LINKED ISSUE #2004: Dark mode tutorial terminal

teemingc:
Fixes https://github.com/sveltejs/svelte.dev/issues/2004

Overrides the `ansi-to-html` colours so that they're more readable in dark mode while maintaining readability in light mode and fixes the dark mode terminal background

Old:
old dark mode terminal

New:
new dark mode terminal

### DIFF
--- a/apps/svelte.dev/src/routes/tutorial/[...slug]/Output.svelte
+++ b/apps/svelte.dev/src/routes/tutorial/[...slug]/Output.svelte
@@ -213,14 +213,7 @@
 		font: var(--sk-font-mono);
 		padding: 1rem;
 		border-top: 1px solid var(--sk-border);
-		background: rgba(255, 255, 255, 0.5);
 		backdrop-filter: blur(3px);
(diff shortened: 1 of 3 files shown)


apps/svelte.dev/src/routes/tutorial/[...slug]/Output.svelte
```diff
@@ -213,14 +213,7 @@
 		font: var(--sk-font-mono);
 		padding: 1rem;
 		border-top: 1px solid var(--sk-border);
-		background: rgba(255, 255, 255, 0.5);
```
Comment (teemingc) [inline apps/svelte.dev/src/routes/tutorial/[...slug]/Output.svelte:216]: We don't need these. The panel itself already has the correct background colour

Rich-Harris (approved):

PR #2019, opened 2026-06-01
fix: update write-less-code repl version from 3 to 5

### LINKED ISSUE #1968: Broken code snippet in blog
Nothing major, more a FYI: https://svelte.dev/blog/write-less-code see the first snippet. 

It loads svelte version 3.9, but it appears to contain $state code.

arieleli01212:
Fixes #1968.

The "write-less-code" blog post embeds a REPL that was migrated to Svelte 5 syntax (using `$state`) in #294, but the iframe's `?version=3` query parameter was not updated, causing the embedded REPL to load Svelte 3 — which doesn't understand `$state` and breaks the snippet.

This changes the version from `3` to `5` so the embedded REPL uses Svelte 5 to compile the example.


arieleli01212:
Closing — a duplicate PR was opened just before this one.

PR #2055, opened 2026-06-23
chore: migrate to `$app/state`

teemingc:
need to do this before https://github.com/sveltejs/kit/pull/15499 to get preview deployments passing

There are two component refactors I'm unsure about that migrate the use of `store.subscribe` to something else entirely


packages/site-kit/src/lib/actions/legacy-details.svelte.ts
```diff
@@ -0,0 +1,65 @@
...
+		unlisten = () => {
+			for (const detail of details) {
+				detail.removeEventListener('toggle', on_toggle);
+			}
+		};
+	}
+
+	// Page changed. Update again
+	$effect(() => {
+		page;
```
Rich-Harris: I haven't tested this locally yet, but I don't think this will do anything? `page` itself never changes, only its properties

dummdidumm: Yeah this needs to listen to a property
```suggestion
		page.url.pathname;
```

dummdidumm (commented):
looks good except the one thing in `legacy_details`

Rich-Harris (approved):

PR #2064, opened 2026-06-26
fix: add `h4` heading styles to design system

sacrosanctic:


Rich-Harris:
The omission of `h4` was deliberate — that level of nesting indicates that content should be refactored. What was the motivation for this? It only shows up in a couple of places, which we should fix

sacrosanctic:
Ok, will ok PR to fix instead.

PR #2071, opened 2026-07-01
chore: add permission to delete preview branches

teemingc:
Adds the `contents: write` permission for the preview branch clean up workflow to avoid these errors https://github.com/sveltejs/svelte.dev/actions/runs/28449714847/job/84308355483#step:3:7

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,1 +1,1 @@
-Run `pnpm prettier --write` on any files you change.
+Run `pnpx prettier --write` on any files you change.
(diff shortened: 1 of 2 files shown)


AGENTS.md
```diff
@@ -1 +1 @@
-Run `pnpm prettier --write` on any files you change.
```
teemingc, line 1: This command fails when run by agents because `prettier` is not installed in the root of the repository. Changing it to `pnpx` at least ensures it work

dummdidumm (approved):

PR #2079, opened 2026-07-05
fix: prevent 404 on llms.txt links after client-side navigation

imlargo:
## Summary
- The `llms.txt` link at the bottom of doc pages (and on `/docs/llms`) points to a route that only has a `+server.ts` (no `+page.svelte`). SvelteKit's client-side router intercepted the click and tried to render it as an SPA page, which fails since there's no page component — resulting in a client-rendered 404. A hard reload bypasses the client router entirely and hits the server route directly, which works correctly.
- Add `data-sveltekit-reload` to these links so they always perform a full browser navigation instead of being intercepted by the SPA router.

## Test plan
- [ ] Visit a docs page (e.g. `/docs/kit/adapter-cloudflare#Usage`) and click the `llms.txt` link at the bottom, confirm it loads the file instead of a 404.
- [ ] Visit `/docs/llms` and click each of the listed `llms*.txt` links, confirm they all load correctly.

### DIFF
--- a/apps/svelte.dev/src/lib/components/PageControls.svelte
+++ b/apps/svelte.dev/src/lib/components/PageControls.svelte
@@ -23,7 +23,7 @@
 		<Icon name="edit" /> Edit this page on GitHub
 	</a>
 	{#if llms}
-		<a href={page.url.pathname.replace(/\/$/, '') + '/llms.txt'}>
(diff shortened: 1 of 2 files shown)


teemingc:
Closing in favour of https://github.com/sveltejs/svelte.dev/pull/2199

PR #2088, opened 2026-07-13
fix: adjust quantity of `../` before $lib alias example

tembra:
It fixes first load on link `/a/deeply/nested/route` before user change the import to use `$lib` alias.


Ocean-OS:
Thank you for making this pull request! However, from what I can tell, the tutorial has the import wrong on purpose; it's meant to show how `$lib` makes it much easier to avoid making these mistakes in your code. In the text for this tutorial, it acknowledges how the import was incorrect (which is why the page initially breaks), and then transitions to demonstrate how `$lib` prevents these issues. 
As such, since this PR kind of contradicts the text/purpose of the tutorial, I'll close this. Thank you, though!

PR #2099, opened 2026-07-18
fix: bump Kit version to include `@sveltejs/kit/env` exports

teemingc:
fixes deployments


apps/svelte.dev/package.json
```diff
@@ -22,7 +22,6 @@
 		"@lezer/javascript": "^1.5.4",
 		"@lezer/lr": "^1.4.8",
 		"@rich_harris/svelte-split-pane": "^3.0.0",
-		"@sveltejs/amp": "^1.1.5",
```
teemingc, line 25: This should be in dev deps because we only use it for the types

PR #2102, opened 2026-07-18
chore: update hooks links

Rich-Harris:
needed once https://github.com/sveltejs/kit/pull/16396 is merged


Rich-Harris:
...err actually never mind, I guess this only applies once the docs are on `main`

PR #2134, opened 2026-08-04
next.svelte.dev

Rich-Harris:
This should not be merged — this branch exists to power next.svelte.dev, once it's set up


Conduitry (changes requested):
Requesting changes so we don't accidentally merge this.

PR #2158, opened 2026-08-11
chore: bump kit

Rich-Harris:
trying to get preview docs deploying


Rich-Harris:
failing with `main` docs because we removed `handleValidationError` — we can probably fix that with an `@errors` annotation if nec

PR #2168, opened 2026-08-12
fix: use a separate package for types resolution

Rich-Harris:
This hopefully fixes, once and for all, the issues we have around type mismatches between the version of SvelteKit (and potentially other dependencies) that are used in the site, and the version that the docs refer to.

### DIFF
--- a/packages/docs-types/package.json
+++ b/packages/docs-types/package.json
@@ -1,5 +1,5 @@
 {
-	"name": "@sveltejs/docs-types",
+	"name": "docs-types",
 	"private": true,
 	"version": "0.0.0",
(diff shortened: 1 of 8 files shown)


packages/docs-types/package.json
```diff
@@ -0,0 +1,26 @@
+{
+	"name": "@sveltejs/docs-types",
```
elliott-with-the-longest-name-on-github: It'd probably be better to not use `@sveltejs/*` for this package name, unless we're actually planning to publish it at some point?

elliott-with-the-longest-name-on-github (approved):

PR #2171, opened 2026-08-13
fix smart quotes

Rich-Harris:
I think this works? Not sure why this broke in the first place, maybe some change around `marked`. Need to spot check a few pages before merging this


teemingc:
Related: https://github.com/sveltejs/svelte.dev/pull/1325

teemingc (approved):
I went through the docs and it's fixed in most places. https://github.com/sveltejs/svelte.dev/issues/781 is a separate issue

PR #2172, opened 2026-08-13
Update conditional class description for clarity

obaibironke:
I changed a word from "truthy" to "true"


Conduitry:
'Truthy' is correct, and 'true' is less accurate. https://developer.mozilla.org/en-US/docs/Glossary/Truthy

PR #2190, opened 2026-08-27
docs: clarify wording in keyed each blocks tutorial

### LINKED ISSUE #2179: Wording choice in Svelte tutorial
In https://svelte.dev/tutorial/svelte/keyed-each-blocks , the tutorial says "By default, updating the value of an each block will add or remove DOM nodes at the end of the block if the size changes, and update the remaining DOM. That might not be what you want."

I think it should be clarified that the "value of an each block" is referring to the iterator the each block is running through. It is a very small change but makes a huge difference in how I understood keyed each block. As a reader, it was not intuitive to me what "value" means, I originally thought it meant a dependent data, kind of like how most runes are executed based on dependent data. I had to find this information myself through clarifying on AI, it took a lot more time to understand this one sub-chapter than necessary, adding this small clarification will help readers immensely.

1t1sCooL:
Fixes #2179

In the keyed each blocks tutorial, the opening sentence says "updating the **value** of an `each` block will add or remove DOM nodes at the end of the block". As reported in #2179, "value" is ambiguous here — a reader can take it to mean some dependent piece of data (the way runes track dependencies) rather than the thing the block iterates over.

This changes the wording to name that thing explicitly:

> By default, updating the array that an `each` block iterates over will add or remove DOM nodes at the _end_ of the block…

One-line change, no behavior or code affected.

### DIFF
--- a/apps/svelte.dev/content/tutorial/01-svelte/04-logic/05-keyed-each-blocks/index.md
+++ b/apps/svelte.dev/content/tutorial/01-svelte/04-logic/05-keyed-each-blocks/index.md
@@ -2,7 +2,7 @@
 title: Keyed each blocks
 ---
 
-By default, updating the value of an `each` block will add or remove DOM nodes at the _end_ of the block if the size changes, and update the remaining DOM. That might not be what you want.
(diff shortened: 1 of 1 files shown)


Rich-Harris:
thank you — already merged #2206 (saw it first) so will close this

PR #2195, opened 2026-08-27
fix: exclude embedded skill content from search index on AI docs pages

### LINKED ISSUE #2061: Search results incorrectly include subsections in AI skills content
I just searched for "inspect.trace" on the built-in search feature on the site, and the first result was "Skills - Overview - $inspect.trace" with a link to https://svelte.dev/docs/ai/skills#$inspect.trace which is not something that currently exists on that page. If you click to expand the skill content for "svelte-code-writer", you do see this, in Markdown form. I think the best way to address this would be to remove these subsections within the skills from the search results.

teemingc:
Fixes #2061

This PR strips `<details>...</details>` content from `docs/ai/*` page bodies before splitting them into searchable sections. The skill headings that are real page sections (e.g. `## svelte-code-writer`) remain indexed, since they live outside the `<details>` blocks.

### DIFF
--- a/apps/svelte.dev/src/routes/content.json/+server.ts
+++ b/apps/svelte.dev/src/routes/content.json/+server.ts
@@ -34,7 +34,13 @@ async function content() {
 		const { slug, body, metadata } = document;
 		const breadcrumbs = document.breadcrumbs.map((x) => clean(x.title));
 
-		const sections = body.trim().split(/^## /m);
+		// skill content embedded in <details> blocks (AI docs) contains its own
+		// headings that don't exist as anchors on the page — exclude from search
+		const searchable_body = slug.startsWith('docs/ai/')
+			? body.replace(/<details>[\s\S]*?<\/details>/g, '')
+			: body;
+
+		const sections = searchable_body.trim().split(/^## /m);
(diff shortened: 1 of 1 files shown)


apps/svelte.dev/src/routes/content.json/+server.ts
```diff
@@ -34,7 +34,13 @@ async function content() {
 		const { slug, body, metadata } = document;
 		const breadcrumbs = document.breadcrumbs.map((x) => clean(x.title));
 
-		const sections = body.trim().split(/^## /m);
+		// skill content embedded in <details> blocks (AI docs) contains its own
+		// headings that don't exist as anchors on the page — exclude from search
+		const searchable_body = slug.startsWith('docs/ai/')
+			? body.replace(/<details>[\s\S]*?<\/details>/g, '')
```
Conduitry:
```suggestion
			? body.replace(/<details>.*?<\/details>/gs, '')
```

I know we're not doing this elsewhere, but I'd like to move toward `s`/`dotAll` as we touch code.

teemingc: TIL!

Conduitry (approved):

PR #2199, opened 2026-08-28
fix: prevent docs routes from intercepting llms.txt links

### LINKED ISSUE #2189: 🐛 llms.txt cannot be found across documentation pages
Image

teemingc:
Clicking an `llms.txt` link from a docs page was incorrectly handled as client-side navigation by the catch-all docs page route, resulting in a rendered 404 even though direct requests reached the text endpoint.

This constrains the docs rest parameter with a matcher that reserves `llms.txt` and `llms-small.txt`, allowing SvelteKit to fall back to normal browser navigation for those server endpoints. 

Closes #2189


apps/svelte.dev/vite.config.ts
```diff
@@ -36,11 +38,13 @@ const plugins: PluginOption[] = [
 	sveltekit({
 		adapter: adapter(),
 
+		files: !is_kit_2 ? { params: 'src/params-registry.ts' } : undefined,
```
teemingc, line 41: Temporary workaround until we remove kit 2 support

apps/svelte.dev/vite.config.ts
```diff
@@ -6,6 +6,8 @@ import { browserslistToTargets } from 'lightningcss';
 import browserslist from 'browserslist';
 import { VERSION } from '@sveltejs/kit';
 
+const is_kit_2 = VERSION[0] === '2';
```
Rich-Harris, line 9: we're using SvelteKit 3 on both `main` and `next` so I think we can get rid of all this?

Rich-Harris: ohhhh wait. we still need it for preview deployments. d'oh

Rich-Harris (approved):

Rich-Harris (approved):

PR #2201, opened 2026-08-28
fix: show Svelte compiler warnings in "on this page"

### LINKED ISSUE #1695: Missing compiler warnings sub nav
https://svelte.dev/docs/svelte/compiler-warnings there are lots of compiler warning types on this page but the sub nav menu doesn't list them at all. It probably should so that it's easier to navigate. It might be because the headings are all h3 instead of h2 but that might have also been a stylistic purpose

teemingc:
Generated compiler warning and error pages use level-three headings without a preceding level-two heading, leaving their "On this page" navigation empty.

This updates the shared docs indexer to promote orphan level-three headings to top-level navigation entries.

Closes #1695

### DIFF
--- a/packages/site-kit/src/lib/server/content/index.ts
+++ b/packages/site-kit/src/lib/server/content/index.ts
@@ -30,6 +30,7 @@ export async function create_index(
 				'<code>$1</code>'
 			);
 
+		let current_section: Section | undefined;
 		const sections = Array.from(body.matchAll(/^#{2,3}\s+(.*)$/gm)).reduce((arr, match) => {
 			if (is_in_code_block(body, match.index || 0)) return arr;
 			const title = match[1];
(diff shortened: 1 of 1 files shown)


packages/site-kit/src/lib/server/content/index.ts
```diff
@@ -40,13 +41,15 @@ export async function create_index(
...
-				}
+			if (match[0].startsWith('###') && current_section) {
+				current_section.subsections.push({
+					slug: `${current_section.slug}-${slug}`,
+					title: displayed_title
+				});
 			} else {
-				arr.push({ slug, title: displayed_title, subsections: [] });
+				const section = { slug, title: displayed_title, subsections: [] };
+				arr.push(section);
```
teemingc, line 52:
```suggestion
				// some pages skip h2 and use h3 such as the svelte compiler warnings
				arr.push(section);
```

Rich-Harris (approved):

Rich-Harris:
FWIW I think we should eventually put every error message on its own page — it would be a lot neater, and it's not like these are pages that people tend to browse. But for now this is much nicer

